PR #2, opened 2017-06-05
Proposed glossary terms

fjrojasgarcia:
I suggest this list of glossary terms to get you started.


aantonop:
Was this written by you or sourced from somewhere? If sourced, I need to add attribution. If from you, I need to add acknowledgment to you!

fjrojasgarcia:
It's a subset (~85%) of `https://github.com/ethereum/homestead-guide/blob/master/source/glossary.rst`, with some terms definition reduced under my personal interpretation and adjusting the alphabetical ordering of all the terms.
I have look for more sources but in my opinion is the best.
I hope it serves as a good starting point.

aantonop:
I'm doing some research to see if the original source (https://github.com/ethereum/homestead-guide/blob/master/source/glossary.rst) has a license that allows derivation and how to properly attribute this

aantonop:
The Homestead Guide is currently without a license. The only proposed PR to add a license is for a CC-BY-SA which would be incompatible with the initial license (CC-BY-NC-ND) of this book. While the book will be released CC-BY-SA after 1 year of publication, it will not at first and therefore it cannot use this glossary.  Unfortunately, I cannot merge this. 

Thanks for the contribution. I guess we will have to build a glossary from scratch

aantonop:
Will ask the primary 2 authors of the original to license their work directly, see if we can use it.

ryanbrownengineer:
Might I suggest crowd sourcing a glossary of confusing/crucial terms, as well as crowd sourcing the definitions? They should become more and more refined as more and more people edit the list.

aantonop:
I have received no response from the authors of this glossary. So, we will start a new one from scratch and crowdsource it. 

Thanks @fjrojasgarcia, but I will have to reject this PR and start a new file for a crowdsourced glossary.

fjrojasgarcia:
I understand and respect your decision @aantonop

PR #81, opened 2017-07-16
Should there be a section on browsers?

ramvi:
In that case I propose putting it in after Parity. This pull requests has a stub text as placeholder


aantonop:
Added manually to clients.asciidoc and dev-tools.asciidoc

Thanks @ramvi

PR #107, opened 2018-01-20
Added missing solution for X=3, Y=17 to the small EC example

krzysztof:
Signed-off-by: Krzysztof Nowak <k.nowak@cern.ch>


krzysztof:
Oops, it was obviously correct, my bad :)

PR #113, opened 2018-01-23
added testnet faucets

ChristopherGondek:
I added some information about test ether faucets. I reasoned why there are faucets and which faucets exist. This list of faucets can, however, never be complete as there may be many different private faucets. I guess we can only link to the most prominent ones.

### DIFF
--- a/ethereum-testnets.asciidoc
+++ b/ethereum-testnets.asciidoc
@@ -8,6 +8,31 @@
 
 === Getting test-ETH - Testnet Faucets
 
+Since testnets do not operate with real money, the incentive to secure the testnets by miners is low.
+Therefore, the testnets must be differently protected against abuse and attacks.
+As a result, so-called faucets exist, which distribute free test ether to developers in a controlled manner.
+
+==== Ropsten
+
+For the Ropsten testnet, different faucets exist:
+
+* http://faucet.ropsten.be:3001/
+This faucet provides the possibility to queue the address that should receive the test ether.
+
+* Metamask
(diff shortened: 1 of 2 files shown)


RichHorrocks:
Hi there. Quick comment: I'm struggling to see the reasoning stating that faucets offer a level of protection against abuse and attacks. Can you add a few clarifying details?

ChristopherGondek:
Hi. Sure. This is the main reason why faucets only give very little ether. Without that restriction, anyone could request lots of ether and constantly sending transactions. That would put the test network under stress and therefore inhibit developers. Also, the blockchain would grow in size unnecessarily. After ethers are distributed, it will be very hard to see which transactions are malicious and which are not. On the other hand, for normal DApp development, few ethers are usually sufficient.

RichHorrocks:
Okay, that makes sense - I hadn't considered that. However, other than the financial disincentive, there's nothing to stop anyone actually mining a PoW testnet, so faucets wouldn't stop someone with some cash to spare from acting maliciously (as was the case with Morden and the first iteration of Ropsten?). Which is then presumably why the PoA testnets were created. (Just thinking out loud... )

ChristopherGondek:
Yes, that is true. However, it requires some resources than to spam a PoW testnet. PoA networks are also highly beneficial when developing DApps. They are more reliable than PoW test networks. They are not as quick as single node networks, but they come with e.g. public blockchain explorers and so on. I personally like to develop DApps on PoA networks in later stages. E.g. to present. This is getting off-topic, however.

PR #125, opened 2018-02-01
Proposed intro chapter changes

krzysztof:
Split into commits for easier cherry-picking.


aantonop:
Wow. This is spectacular review and editing work. Thanks so much!

PR #141, opened 2018-02-14
Found some typos

fjrojasgarcia:
I have found some typos.

I would also like to inform that I have found also some broken links, "https://github.com/ethereumbook/ethereumbook/blob/first_edition/code/METoken", "https://github.com/ethereumbook/ethereumbook/blob/first_edition/code/METoken/truffle.js", and "https://github.com/ethereumbook/ethereumbook/blob/first_edition/code/METoken/contracts/METoken.sol".
*broken

About truffle project initialization, I personally don't know what have to do the link "https://github.com/ethereumbook/ethereumbook/blob/first_edition/code/Faucet.sol " in this section.


aantonop:
The links to first_edition will exist when I merge into that branch. I want all the links in the book to point to a specific branch or tag, rather than develop@HEAD. That way we can know that the examples in the book are consistent with each other at the time of publication. That's why those links are broken, the first_edition tag doesn't exist yet.

PR #169, opened 2018-02-23
Add missing definition for ommer/uncle

zmxv:


glossary.asciidoc
```diff
@@ -111,7 +111,8 @@ network::
 node::
     A software client that is participating in the peer-to-peer network.
 
-ommers::
+ommer (a genderqueer term for "uncle/aunt")::
```
mark-s: How about 'gender neutral'  ?

aantonop: I would agree, ommer is gender neutral, not gender queer. It has been adopted in gender queer language, but came from programming AFAIK

PR #182, opened 2018-02-26
Smart Contract intro and definition to the top

diegogurpegui:
Move the Smart Contract intro and definition to the top, and start drafting it.
I believe it's best to start answering What is a Smart Contract before going deep into to the different languanges. The explanation should be a bit more detailed that the one given in the introductory chapters

### DIFF
--- a/smart-contracts.asciidoc
+++ b/smart-contracts.asciidoc
@@ -1,5 +1,11 @@
 == Smart Contracts
 
+=== What is a Smart Contract?
+
+A Smart Contract is a computer logic that, given certain conditions, can enforce the execution of one or more transactions in an irreversible and decentralized way without the need for third parties. The term was first used by Nick Szabo in 1994.
+
+Bitcoin supports smart contracts in a limited way by _Turing incomplete_ Script languange (using what's called OP CODES). On the other hand, the Ethereum blockchain uses Turing complete languages which makes the Smart Contracts essentially programs that run in a distributed way. In that sense, they are pieces of code that live on the blockchain and execute commands exactly how they were told to. They can read other contracts, make decisions, send ether and execute other contracts. Contracts will exist and run as long as the whole network exists, and will only stop if they run out of _gas_ or if they were programmed to self destruct.
+
 [[high_level_languages]]
 === Introduction to Ethereum High-Level Languages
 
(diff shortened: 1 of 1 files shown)


smart-contracts.asciidoc
```diff
@@ -1,5 +1,11 @@
 == Smart Contracts
 
+=== What is a Smart Contract?
+
+A Smart Contract is a computer logic that, given certain conditions, can enforce the execution of one or more transactions in an irreversible and decentralized way without the need for third parties. The term was first used by Nick Szabo in 1994.
+
+Bitcoin supports smart contracts in a limited way by _Turing incomplete_ Script languange (using what's called OP CODES). On the other hand, the Ethereum blockchain uses Turing complete languages which makes the Smart Contracts essentially programs that run in a distributed way. In that sense, they are pieces of code that live on the blockchain and execute commands exactly how they were told to. They can read other contracts, make decisions, send ether and execute other contracts. Contracts will exist and run as long as the whole network exists, and will only stop if they run out of _gas_ or if they were programmed to self destruct.
+
```
mark-s: Not sure about the level of rigour you're after here but; the Ethereum EVM also uses OPCODES (the concept may be needed when talking about gas). And it's not really the languages that provide the Turing completeness - but instead the EVM has a Turing complete instruction set the languages can make use of. Also, I'm not sure it's correct to say that '[contracts] can read other contracts' as this implies they can access the source code?
Contracts only execute for a brief time, here it sounds as if they are some kind of background service ('run as long as the whole network exists').

Again, apologies if this is a bit nit-picky.

aantonop: An important fact that needs to be highlighted early:

Contracts run when a transaction triggers them, and ONLY in response to a transaction. All contract execution starts with a transaction signed by an EOA. Contracts do not run "on their own" ever.

diegogurpegui: You are both correct. I took most of that definition from the official Ethereum wiki just to start drafting the content of that section. I'll adjust it.

Regarding the OPCODE, I can remove it from there since it might add more confusion that clarification.

Benudek:
I suggest 

1. compare this to a Java program or database stored procedure in a table what is different what is equal: the shared infrastructure and hence immutability is different. The crypto gas part is different 

2. discuss the term smart contract. It seems these programs are not smarter than any other program. There are some similarities to contracts but naturally also clear differences 

Summarized: demystifying the metaphor and comparing to existing computer programs and the infrastructure they run on would be helpful in my humble opinion

PR #185, opened 2018-02-26
real-Wallet Update glossary.asciidoc

EvertonMelo:
font: https://etherscancom.freshdesk.com/support/solutions/articles/35000022140-what-is-etherscan-


aantonop:
It is not necessary to store public keys - they can be derived from the secret keys quickly.

PR #187, opened 2018-02-27
Patch 1

h4sh3d:
Added missing prefixes in compressed point table, and added resistance to first and second pre-image.

I think a more clear explanation/separation between the format used to create addresses (64 bytes) and the standard ways to format point (33 bytes or 65 bytes) should be done. It is confusing with the sentence "Ethereum only uses uncompressed public keys..."


aantonop:
Nice work!

Agree on the need to be more specific in the description of compressed/uncompressed, especially since Ethereum _doesn't really use public keys at all_. Unlike bitcoin that stores them in UTXO scripts, in Ethereum they are only used as pre-images to addresses. Great point!

Would you like to propose some wording to capture that nuance? Or should I add it as an issue and I can tackle it at a later time?

h4sh3d:
I'll think about it but I let you create an issue to keep track of it. Maybe other people will propose good wording to capture the nuance.

PR #189, opened 2018-02-28
terms gas.asciidoc

EvertonMelo:
reference documentation: 

http://ethdocs.org/en/latest/ether.html?highlight=gas

### DIFF
--- a/gas.asciidoc
+++ b/gas.asciidoc
@@ -2,6 +2,9 @@
 == Questions that need to be answered in this section
 
 * What is Gas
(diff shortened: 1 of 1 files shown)


aantonop:
Cannot use this source, license is not suitable. Please see:
https://github.com/ethereumbook/ethereumbook/blob/develop/CONTRIBUTE.md#license-and-attribution

If you would like to suggest sources, please add the link, rather than copying text. But please make sure the license is compatible (MIT, CC-BY). 

Thank you for the effort!

PR #200, opened 2018-03-02
Oracles

timnugent:
First stab at an Oracles section - probably to end up within the Dapps section


aantonop:
Fantastic work Tim!

I will move the chapter into the contrib/ folder so that it can be merged into the main chapters after review and edits.

fjrojasgarcia (commented):
Awesome chapter!

PR #211, opened 2018-03-04
Additions to the glossary - Library in Solidity, Inmutable Deployed Code Problem, Paralysis Problem and Paralysis Proof System

fjrojasgarcia:
I propose this two terms/concepts to the glossary;

**Library in Solidity** and **Inmutable Deployed Code Problem**


fjrojasgarcia:
Added terms Paralysis Problem and Paralysis Proof System

PR #215, opened 2018-03-05
gas: updated information on who pays for gas

nichanank:
added information on who pays for gas
also added my name to contributors list


aantonop:
Really good work Nichanan! Thank you

PR #220, opened 2018-03-07
Add instructions for using puppeth on GCP for a private testnet

jaycenhorton:
This PR gives instructions necessary for using Puppeth (a standard geth component) to deploy a private ethereum testnet with miners, a wallet, a block explorer, ethstats, and connection instructions. I am submitting it in a pretty raw format. If you find this useful, we can offer a cleaner version.

### DIFF
--- a/ethereum-testnets.asciidoc
+++ b/ethereum-testnets.asciidoc
@@ -72,16 +72,24 @@ https://rinkeby.etherscan.io/
 
 === Launching a testnet on AWS or Azure
 
-=== Launching a testnet on Google Cloud Platform using (puppeth!)[https://github.com/ethereum/go-ethereum/tree/master/cmd/puppeth]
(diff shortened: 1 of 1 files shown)


stefek99:
I'm just thinking that VPN business is adding an extra layer of complexity.

If I was following a tutorial on Ethereum private testnets, I'd rather spend my energy dealing with Ethereum related issues.

But maybe it's just me, I was never a big fan of configuring stuff. Happier to download 10GB VM with the software that just works than spend whole working day debugging not-compiling dependencies.

**UPDATE:** https://github.com/ethereum/go-ethereum/wiki/Private-network

aantonop:
This will be moved into the "contrib" directory, as it is a rather large contribution that will probably go in an appendix.

Thank you!

PR #245, opened 2018-03-11
Explains the gas price and limit in image shown in introduction.

EdgarBarrantes:


come-maiz (approved):
Thank you! I was very confused about this, because couldn't figure out the 63 from the numbers presented. Big +1 here.

PR #249, opened 2018-03-11
Fixed typos

dimitris-t:
"how those properties makes them so useful" -> "how those properties make them so useful"
"descirbed" -> "described"


dimitris-t:
I can squash all of those commits into one.

PR #254, opened 2018-03-13
Capitalized Bitcoin

ecurrencyhodler:
If Ethereum is capitalized, so too should Bitcoin as they are both proper nouns.


ecurrencyhodler:
I see now you haven't capitalized ether and understand now how you have distinguished ether from Ethereum.  Apologies.

PR #259, opened 2018-03-13
AWS set-up; description of topolgy.

chainhead:
Hi Andreas,
I just finished the basic topology description of AWS set-up. Next up is `bootnode` set-up and deploying the MET on to this network.

Regards,
Nagesh


aantonop:
Excellent work. 

If I may suggest, it would be helpful for each of the machines to have names that are role-based throughout the document. Rather than referring to "Ethereum-0" or the "first Ethereum node", instead use role-based names such as "mining node", "rpc node" "client". Use these in the hostname, configuration file, the text descriptions, the diagrams, everywhere. It will make it a lot easier to read and comprehend the layout

chainhead:
Hi Andreas,
Yes, those were supposed to be place-holders to be changed with 'Find all, replace all'. I will do that in the final version.

PR #260, opened 2018-03-14
Creating Vyper information

tpmccallum:
There are so many changes (between Solidity and Vyper), this section will provide an overview of each area as well as sample code and justification of the changes.


aantonop:
Thank you!

I will be moving this file to the contrib/ directory so that I can put it in the pipeline for editing. Continue any further PR's with the new path.

PR #263, opened 2018-03-14
Some corrections concerning opcodes and Solidity.

chriseth:


evm.asciidoc
```diff
@@ -100,21 +100,21 @@ Let's examine the first two instructions:
 PUSH1 0x60 PUSH1 0x40
 ----
 
-Here we have the _mnemonic_ "PUSH1" with a _value_ of "0x60". This corresponds to the EVM instruction of pushing a 1-byte item onto the stack. This operation is capable of pushing a 32-byte item on the stack with the following opcode:
+Here we have the _mnemonic_ "PUSH1" followed with a raw byte of value "0x60". This corresponds to the EVM instruction of interpreting the single byte following the opcode as a literal value and pushing it onto the stack. It is possible to push values of size up to 32 bytes onto the stack. For example, the following bytecode pushes a 4 byte value onto the stack:
```
chriseth, line 103: I tried to make it a little clearer, but in general, the different push instructions are hard to explain without looking at the raw bytecode.

PR #278, opened 2018-03-15
Smart contracts

BrianGuo:
Added paragraph into the best-practices section of the Smart-Contracts asciidoc describing security and gas. This is a preliminary pull request, opening up for comments and suggestions about what to add/remove/fix. We also improved upon the introductory paragraph on smart contracts, removing some of the more informal language.


aantonop:
Nice work @BrianGuo

PR #283, opened 2018-03-16
Updated smart contract deployment

GPacenza:


aantonop:
This seems to be duplicating much of the work already covered in:
 dev-tools.asciidoc
clients.asciidoc and
ethereum-testnets.asciidoc

Is there a reason we need to repeat all this? We've demonstrated it with truffle. Do we need to also show it using geth? If yes, wouldn't that fit better in dev-tools or clients?

GPacenza:
So the idea was to focus more on deploying the contract itself and less on the actual technology - for instance, the geth section in clients.asciidoc seems to be focused on getting started with the geth client but doesn't actually go over deployment of a contract with it. I was hoping to address that issue. As far as repetition, I felt it was worthwhile to go over this again with geth despite the fact it was demonstrated with Truffle just to present the reader with multiple tools to use. Again, it was placed in smart contracts because the idea was to focus more on the deployment of the contract and how to do it, not so much on the tools themselves.

But just to clarify, were you referring only to the geth section or to the section on Infura as well? And with something like smart contract deployment, do you have any advice or ideas on how to differentiate between what should go in smart contracts and what should go in dev tools?

smart-contracts.asciidoc
```diff
@@ -87,6 +87,117 @@ The link includes details about the formal specification of encoding and various
...
+
+instance: Geth/v1.8.1-stable/darwin-amd64/go1.10
+INFO [03-14|18:34:37] Etherbase automatically configured       address=0x6e6A1eFF05ba3a16c3A3E5a274B288b10490C428
+coinbase: 0x6e6a1eff05ba3a16c3a3e5a274b288b10490c428
+at block: 4535991 (Sat, 11 Nov 2017 21:17:37 EST)
+ datadir: /Users/brianleffew/Library/Ethereum
+ modules: admin:1.0 debug:1.0 eth:1.0 miner:1.0 net:1.0 personal:1.0 rpc:1.0 txpool:1.0 web3:1.0
+----
+
+If you this is your first time using geth, it might take a while to sync up to the network.
```
come-maiz, line 136: remove "you": If _ this is [...]

PR #299, opened 2018-03-19
Update info about regaining access to account

AyushyaChitransh:
Users should be made aware of the fact that _only_ password cannot help them regain access to their account. They need to have "keystore" file also.

Also, users should also know which things can be used in case their system undergoes a format. This is the most general cause of losing access to acccount.


fjrojasgarcia (approved):
Ship It!

PR #305, opened 2018-03-20
[fix] Unclosed table in smart-contracts asciidoc

BrianGuo:
Also fixed unclosed table in smart-contracts

Note: the formatter that I was using automatically removed unnecessary spaces from the end of lines. The only line I actually changed was the unclosed block, line 243, which caused formatting issues in the rest of the doc


fjrojasgarcia (approved):
IMHO It seems a clean and a not so invasive auto formatting work.

Ship It!

PR #308, opened 2018-03-21
Elaborate on nonce

AyushyaChitransh:
There are two types of nonce used in ethereum, explained in short about them in glossary.


fjrojasgarcia:
Totally agree. IMHO Approved.

aantonop:
@fjrojasgarcia Feel free to merge immediately, yourself, if you think it is ready.

PR #310, opened 2018-03-21
Some restructuring of the article and ETC Testnets

pyskell:
Restructured a bit so that general information on the purpose of testnets is better separated from the details of connecting to/using them.

Also added ETC testnet details.

I'm wondering too if we should adapt this so that each testnet has the type of template I specified for ETC, this would cleanup the `History of Ethereum Testnets` section quite a bit as at the moment it seems to contain various links that may be better suited to their specific sections.

### DIFF
--- a/ethereum-testnets.asciidoc
+++ b/ethereum-testnets.asciidoc
@@ -63,6 +63,15 @@ https://www.rinkeby.io/
 
 https://rinkeby.etherscan.io/
 
+== Ethereum Classic
+
+=== Morden
+Ethereum Classic currently runs a variant of the Morden testnet that is kept at feature parity with the Ethereum Classic live network. You can connect to it through either the gastracker RPC or by providing a flag to `geth` or `parity`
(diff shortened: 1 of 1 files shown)


ethereum-testnets.asciidoc
```diff
@@ -1,9 +1,8 @@
-== Ethereum Testnets
+== Testnet Overview
+=== What is a testnet?
 
-=== What is testnet?
-
-A test network is used to simulate the behavior of the main ethereum network. There are some publicly available test networks.
-When any major change is about to be included in the main ethereum network, its tests are mostly done on these test networks.
+A test network is used to simulate the behavior of a main network. There are some publicly available test networks.
+When any major change is about to be included in the a main network, its tests are mostly done on these test networks.
```
come-maiz: typo: the a main

pyskell:
Fixed the typo, thanks @elopio !

PR #316, opened 2018-03-22
some review correction for the transaction ECDSA chapter.

LiangMa:
correct the congruence modulo symbol(original is "=")
also replace the confused  p symbol express 2 different concept
add the "mode p" for the verification phase


aantonop:
I agree that the use of p created ambiguity. But the use of k also creates ambiguity, as it is already used for the signing key. I'd like to stick to the pattern of using lowercase for private key, upper case for public. Instead of p,P let's use q,Q for the ephemeral key. k,K for the user signing key. R for the x-coordinate. p remains the module order.

Also, for the congruence module symbol you will have to use an ampersand unicode so that it renders in asciidoc correctly. See how we treated the Xi and rombus characters in the intro section.

aantonop:
Super! Thanks!

PR #341, opened 2018-03-25
Fixed some comma and wording errors in the Transactions section

sgvictorino:


fjrojasgarcia:
Same for this PR.

I do not merge the changes since my English is not native but from my part it is approved.
In general I think you do a good and useful job of proofreading.

I prefer to have the @aantonop's criterion

PR #342, opened 2018-03-25
Added Gas Anchor Heading

sgvictorino:
Am I doing this right @aantonop?
See the link here:
https://github.com/ethereumbook/ethereumbook/blob/develop/smart-contracts.asciidoc#gas

### DIFF
--- a/gas.asciidoc
+++ b/gas.asciidoc
@@ -1,4 +1,4 @@
-[[Gas]]
+[[gas]]
 == Gas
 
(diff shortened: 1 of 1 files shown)


gas.asciidoc
```diff
@@ -1,4 +1,4 @@
-[[Gas]]
+[[gas]]
```
sgvictorino: I've opened #346, assuming these are case sensitive.

PR #351, opened 2018-03-26
what-is: fixed typo

come-maiz:
cost -> costed


aantonop:
Sorry, past tense of "cost" is "cost". I think "costed" is an Americanism and my British mother who taught me "proper" English would not accept that ;-)

PR #370, opened 2018-03-30
clients: cleanup and new JSON-RPC section

aantonop:
- Cleanup and language fixes
- Headings moved down one level
- Added JSON-RPC section with examples
- Added Parity Geth Compatibility mode section


clients.asciidoc
```diff
@@ -1,32 +1,33 @@
...
 
 However, to run a full node you will incur a significant cost in hardware resources and bandwidth. A full node must download more than 50GB of data (as of Feb 2018) and store it on a local hard drive. This data burden increases quite rapidly every day as new transactions and blocks are added. More on this topic in <<requirements>>.
 
-A full node in the _mainnet_ is not necessary for Ethereum development. You can do everything you need to do with a _testnet_ node (which stores a copy of the smaller public test blockchain), or with a local private blockchain.
+A full node running on the Ethereum Main Network (mainnet) is not necessary for Ethereum development. You can do everything you need to do with a _testnet_ node (which stores a copy of the smaller public test blockchain), or with a local private blockchain. You can also develop Ethereum applications with a local private blockchain (see <<ganache>>), or use a cloud-based Ethereum client offered by a service provider (see <<infura>>).
```
fjrojasgarcia: To be precise, here I'd say, _"You can do almost everything you need"_ ...

Taking into account that the **state** (e.g.  a certain state of a/some contract(s) or the previous existence of certain libraries) of the system could be crucial to perform certain tests, it would be possible to find the case in which **the state** with which to perform a set of tests **did not exist in the Testnet** network or to reach that state would require a considerable effort.

clients.asciidoc
```diff
@@ -1,32 +1,33 @@
...
 
-A lightweight client can be used to connect to existing networks, such as your own full node, the public Ethereum blockchain, the public testnet, or a local test blockchain. In practice, you will likely use a lightweight client such as MetaMask as a convenient way to switch between all of the different node options.
+
+A lightweight client is an Ethereum client that does not store a local copy of the blockchain, or validate blocks and transactions. It offers the functions of a wallet and can create and broadcast transactions. A lightweight client can be used to connect to existing networks, such as your own full node, the public Ethereum blockchain, the public testnet, or a local test blockchain. In practice, you will likely use a lightweight client such as MetaMask as a convenient way to switch between all of the different node options. The terms "lightweight client" and "wallet" are used interchangeably, though there are some differences. Usually, a lightweight client offers an API (such as the web3js API), in addition to the function of a wallet.
```
fjrojasgarcia: To try to avoid possible confusions to the very new comers, perhaps here it would be good to highlight the difference of this type of wallets in Ethereum with respect to **SPV Wallets** in **UTXO models**.

Maybe a comment as simple as, "Do not confuse with ..."

clients.asciidoc
```diff
@@ -310,18 +316,60 @@ $
...
+{"jsonrpc":"2.0","id":1,"result":"Geth/v1.8.0-unstable-02aeb3d7/linux-amd64/go1.8.3"}
+----
+
+In this example, we use +curl+ to make an HTTP connection to address +http://localhost:8545+. We are already running +geth+ and it is offering the JSON-RPC API as an HTTP service on port 8545. We instruct +curl+ to use the HTTP +POST+ command, and to identify the content as +Content-Type: application/json+. Finally, we pass a JSON-encoded request as the +data+ component of our HTTP request. Most of our command line is just setting up +curl+ to make the HTTP connection correctly. The interesting part is the actual JSON-RPC command we issue:
+
+----
+{"jsonrpc":"2.0","method":"web3_clientVersion","params":[],"id":1}
+----
+
+This request says that we want to use the +2.0+ version of the JSON-RPC, to call the method +web3_clientVersion+ with no parameters and with network id +1+ (the Ethereum mainnet).
```
fjrojasgarcia: According with [JSON-RPC 2.0 Specification](http://www.jsonrpc.org/specification), an rpc call is represented by sending a Request object to a Server with the following members:

1. **jsonrpc**: Version of the JSON-RPC protocol. Here MUST be exactly "2.0".
2. **method**: The name of the method to be invoked.
3. **params**: A structured value that holds the parameter values to be used during the invocation of the method. This member MAY be omitted.
4. **id**: An identifier **established by the Client** that MUST contain a String, Number, or NULL value if included. The Server MUST reply with the same value in the Response object if included. This member is used to correlate the context between the two objects.

If, for example, you use the following rpc call with **id=1 and id=4**, it comes to the same result because the **network id** is determined by the **endpoint** not the id member of the rpc call object;


```
$ curl -X POST --data '{"jsonrpc":"2.0","method":"eth_blockNumber","params":[],"id":1}' -H "Content-Type: application/json" https://mainnet.infura.io
$ {"jsonrpc":"2.0","id":1,"result":"0x51c076"}`

$ curl -X POST --data '{"jsonrpc":"2.0","method":"eth_blockNumber","params":[],"id":4}' -H "Content-Type: application/json" https://mainnet.infura.io
$ {"jsonrpc":"2.0","id":4,"result":"0x51c076"}`

```

clients.asciidoc
```diff
@@ -310,18 +316,60 @@ $
...
+----
+$ curl -X POST -H "Content-Type: application/json" --data '{"jsonrpc":"2.0","method":"eth_gasPrice","params":[],"id":1}' http://localhost:8545
+
+{"jsonrpc":"2.0","id":1,"result":"0x430e23400"}
 ----
 
-////
-TODO
-////
+The response, +0x430e23400+, tells us that the current gas price on mainnet is 1.8 Gwei (gigawei or billion wei).
```
fjrojasgarcia: Again here, I disagree with the use of the rpc request object member "id", trying to apply it to the identification of the network.

This **"id"** member should be used to correlate the context between the request and replay rpc objects and the input parameters for the method to call should be supplied with the member "params" of the rpc call request object.

fjrojasgarcia (changes requested):

aantonop:
Great catch. I misunderstood the purpose of ID

>

clients.asciidoc
aantonop: Added @50504cd

aantonop:
Merge got messed up so I manually merged after addressing all the comments

PR #381, opened 2018-04-02
Vyper’s compiler responsibilities i.e. opcode sync

tpmccallum:


contrib/vyper.asciidoc
```diff
@@ -66,3 +66,6 @@ While smart contracts can write to the Ethereum's chain data (through log events
...
 
+=== OPCODES
+The code for smart contracts is mainly written in high level languages like Solidity and now potentially Vyper. The compiler is responsible for taking the high level code and creating the lower level interpretation of the code, which is then able to be executed on the Ethereum Virtual Machine (EVM). The lowest representation which the compiler can distille the code to (prior to execution by the EVM) is opcodes. This being the case, each implementation of a high level language (like Vyper) is required to provide an appropriate compilation mechanism (a compiler) to allow (among other things) the high level code to be compiled into the universally predetermined and predefined EVM opcodes. A good example of this is Vyper's implementation of Ethereum's sharding opcodes. TODO I will provide a concrete example here but require a concrete resolution to the following submitted Vyper issue < https://github.com/ethereum/vyper/issues/745 > which specifically relates to the implementation of sharding opcodes in Vyper. This will make for a great example because it is future focused.
```
fjrojasgarcia, line 70: I see this PR correct, however, I think that as a general optimization area, we could indicate the DRAFT status of the document following the recommended practices.

I'd suggest at least indicating DRAFT status as following;

```
= Title of the document: This is the subtitle
:revnumber: v1.1
:revdate: 15.05.2017
:revremark: First draft
```
I'd suggest to adopt this practice or another more optimal, among the participants of the repository. I would do so, at least while there are parts of the document pending development.

fjrojasgarcia (approved):

contrib/vyper.asciidoc
tpmccallum: Hi @fjrojasgarcia 
Sorry about the delay in responding, would you recommend having this draft status example,  once only, at the top of a document (indicating which of its subsections are under construction/draft) or would you recommend having the draft status at the top of every section which is in draft i.e. above the ===OPCODES section (in this case)?
I think that having this practice will also provide a mechanism for authors to search the repository for draft work which requires completion/review. For example a contributing author could search the repository for "revnumber" (or perhaps we could add an additional flag for search, just thoughts).
Many thanks
Tim

fjrojasgarcia: I'd say the latter the better but more elaborated. Perhaps, as a minimun I'd recommend it at document level and use the TODO tags along side the doc could allow us to search pending points.

tpmccallum:
Hi @fjrojasgarcia 
Ok sounds great. I will use the latter (marking each section which requires more research/writing). In addition to this I will use the TODO tag to explicitly to set down (and make searchable) the TODO tasks.

PR #383, opened 2018-04-02
Change Proof-of-State to Proof-of-Stake

tilmans:
Small fix in two places, does not occur in the other documents.


fjrojasgarcia:
Good catch! Thanks for your contribution @tilmans

Issue #85, opened 2017-07-17
Privacy on ethereum


fjrojasgarcia:
I think Privacy on ethereum could be an important aspect to be cover by the book.

Thre are already several projects trying to implement privacy on ethereum, most notably Zcash-on-ethereum (ZoE), Hawk and the later Zcoin.

QwQiao:
Agreed. The book could dedicate a section on zkSNARKs that went live with Byzantium.

fjrojasgarcia:
ZK-SNARKs have a newer, shinier cousin: **ZK-STARKs**. With the T standing for “transparent”, ZK-STARKs resolve one of the primary weaknesses of ZK-SNARKs, its reliance on a “trusted setup”. 

https://vitalik.ca/general/2017/11/09/starks_part_1.html

swasilyev:
> ZK-SNARKs have a newer, shinier cousin: ZK-STARKs. With the T standing for “transparent”, ZK-STARKs resolve one of the primary weaknesses of ZK-SNARKs, its reliance on a “trusted setup”.

Unfortunately that is not perfectly true. Though STARKS solve trusted setup problem, as some other nearly-practical solutions, see the list [https://zkp.science], I don’t think they are practical for Ethereum, at least until sharding/TrueBit/etc. Trusted setup is paid by SNARKS for constant-size proofs, and this price may appear inevitable, as I fail imagine prooving linear computation in constant time. Anyway, since Metropolis, that made them practical, we haven’t seen any reasonable applications, and STARKS can take even more time for adoption.

Regarding book content, the subject is complex, explanation to the depth where it becomes practical for non-specialist would take much space and may feel frightening to vast audience. More feasible may be a brief description of ideas behind the related Metropolis changes, and the possibilities that it created. The hard problem is the selection of topics as we  have no prooven approaches at the moment. It should be at least some primitives from “classical” cryptography, like Pedersen commitments, ring signatures, etc and at least SNARKS from zk-zoo, but it’s hard to which limits.

It’s all too opinionated for a single person at the moment, but if we can make it a collective effort I would be happy to elaborate.

jeamick:
Hello there , for privacy on Ethereum we can talk about Multi party  computation with reference to the Enigma Project  : https://github.com/enigmampc

PR #391, opened 2018-04-05
privacy enhancing patterns

SylTi:
What do you guys think?
Any other patterns in minds? @aantonop 
#85

### DIFF
--- a/preface.asciidoc
+++ b/preface.asciidoc
@@ -187,3 +187,4 @@ Following is an alphabetically sorted list of notable GitHub contributors, inclu
 * Yeramin Santana (ysfdev)
 * Zhen Wang (zmxv)
 * ztz (zt2)
+* Sylvain Tissier (SylTi)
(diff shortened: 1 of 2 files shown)


preface.asciidoc
```diff
@@ -187,3 +187,4 @@ Following is an alphabetically sorted list of notable GitHub contributors, inclu
 * Yeramin Santana (ysfdev)
 * Zhen Wang (zmxv)
 * ztz (zt2)
+* Sylvain Tissier (SylTi)
```
zaq1tomo, line 190: I think these names sorted by alphabet.

aantonop: Don't worry, I'll resort them from time to time. It is better to add to the end for now, so that I can track merges easier.

aantonop:
This is an interesting pattern. I'd like some more people to review the contracts themselves before I merge. Comments please?

SylTi:
I have some implementations tests in JS if you're interested.

edit: Also the `PrivacyJudge` is NOT production ready, for example, if user2 never deposit funds inside the judge, user1 funds are locked forever (but this can easily be solved by adding conditions). They are probably other attacks scenarios possibles.

PR #398, opened 2018-04-06
Adding pending tasks metadata

tpmccallum:
Vyper is still under heavy development and as such there are a few areas which I would like to keep an eye on (rather than writing about redundant or deprecated information). 
I have added revision number, revision date, revision remark and revision status to keep track of the research and development of each Vyper subsection.
These will all be updated with the latest available information in time for ethereumbook's publication.
I have also added some formatting to make the text neater.


tpmccallum:
Hi @fjrojasgarcia
I have implemented the pending tasks metadata as suggested. I will work on this over the weekend. 
Kind regards
Tim

fjrojasgarcia:
Thanks for keep this into account.

Issue #279, opened 2018-03-15
Ensure all headings have asciidoc reference anchors and all references point somewhere


aantonop:
I could use some extra help with this issue. 

In asciidoc, every chapter heading, section heading, figure, table, and code example, can be referenced from anywhere in the book. To do that, each of the above must have an _anchor_, which is a unique name inside double square brackets like [[anchor]]. See this example below:

```
[[transaction_chapter]]
== Transactions

[[transaction_example]]
.An example of a transaction in hex
----
453425235325345
----

[[metamask_transaction_example]]
.A screenshot of MetaMask doing a transaction
image::images/metamask_screenshot.png[A screenshot of Metamask]
```

It doesn't matter what the variable name inside the [[]] is, as long as it is unique. I prefer to use descriptive names with underscore between words, up to 2-4 words in length. 

Elsewhere in the document you can refer back (or forwards) to a section, example, table or figure, by including the reference name inside angle brackets, like this:

```
As we saw in example <<transaction_example>>, a transaction is created 
and then Metamask shows you an approval screen (see <<metamask_transaction_example>>). 

```

When rendered by O'Reilly for print these are replaced by the heading, title, or caption. When rendered in an ebook or HTML they also become hyperlinks to the section. 

Help me make it all work by doing the following:

* A pull request adding anchors to every section, heading, image, code example.
* If there are references that don't point anywhere like <<this_doesnt_exist_yet>>, then go and add a [[this_doesnt_exist_yet]] somewhere else in the text where it makes sense and points to something useful, like a section heading in another chapter.
* Create an issue if you find a reference that doesn't have a good place to point to and is left "dangling"

abhiShandy:
is there a need to anchor TIPS and WARNINGS?

aantonop:
No

nuke-web3:
@aantonop - close #279 ? and/or add this to contributions page for future reference?

PR #408, opened 2018-04-07
Add instructions for interactig with geth testnets

amanusk:
This adds some basic instruction to interacting with available testnets in geth

### DIFF
--- a/clients.asciidoc
+++ b/clients.asciidoc
@@ -155,7 +155,7 @@ For other operating systems, use the package manager of your OS or follow the Wi
 
 Now you have +git+, +golang+, +rust+, and necessary libraries installed, let's get to work!
 
-[[Go-Ethereum]]
+[[go_ethereum_geth]]
 ==== Go-Ethereum (Geth)
 
 There are three original implementations of the Ethereum client written in three different languages: C++, Python, and Go. Geth is the Go language implementation, which is actively developed and considered the "official" implementation of the Ethereum client. There is some debate on the meaning of the title "official client" in a decentralized system such as Ethereum, but suffice it to say that Geth is supported by the Ethereum Foundation, a Swiss non-profit organization founded by Ethereum's creator, Vitalik Buterin.
--- a/clients.asciidoc
+++ b/clients.asciidoc
@@ -155,6 +155,7 @@ For other operating systems, use the package manager of your OS or follow the Wi
 
 Now you have +git+, +golang+, +rust+, and necessary libraries installed, let's get to work!
 
+[[Go-Ethereum]]
 ==== Go-Ethereum (Geth)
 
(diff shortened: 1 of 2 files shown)


clients.asciidoc
```diff
@@ -155,6 +155,7 @@ For other operating systems, use the package manager of your OS or follow the Wi
 
 Now you have +git+, +golang+, +rust+, and necessary libraries installed, let's get to work!
 
+[[Go-Ethereum]]
```
fjrojasgarcia: Andreas point out in #279, _"I prefer to use descriptive names with **underscore** between words, up to 2-4 words in length"_.
IMHO, I'd also avoid letter capitalization.

ethereum-testnets.asciidoc
```diff
@@ -23,7 +23,40 @@ Infura was born with the goal of delivering stable and reliable RPC access to th
 https://infura.io/docs/gettingStarted/chooseaNetwork
 
 ==== Geth
-//// TO DO: Loosely explain the geth commands to connect to a testnet
+Geth natively supports both the Ropsten and Rinkeby networks. To connect to the Ropsten network use the command-line argument:
+----
+geth --testnet
+----
+This will start syncing the Ropsten blockchain. A new directory named 'testnet' will be created in you main Ethereum data directory. A 'keystore' directory will be create inside 'testnet' and will store the private keys of you testnet accounts. As of writing this, the Ropsten blockchain is significantly smaller than the main Ethereum blockchain, and consists of about 14GB of data. Since the testnet requires less resouces, it is simpler to setup and test your code on the testnet first.
```
fjrojasgarcia: "of you testnet accounts" -> "of **your** testnet accounts"

ethereum-testnets.asciidoc
```diff
@@ -23,7 +23,40 @@ Infura was born with the goal of delivering stable and reliable RPC access to th
...
+
+More on setting up geth at: <<clients.asciidoc#Go-Ethereum,Go-Ethereum (Geth)>>
+
+[TIP]
+====
+Although much smaller, it might still take time for the testnet to fully sync. You can check if syncing was complete by running the following command in the geth intractive console:
+----
+eth.getBlock("latest").number
+----
+This should return a number other than 0 once the your testnet node is fully in sync.
```
fjrojasgarcia: "once **the your** testnet node" -> " once your testnet node"

fjrojasgarcia: The fact that the last block number != 0 is not enough to conclude that your node is fully synced.
For instance, you could look for the last block number of the network in a known explorer and compare it.

fjrojasgarcia (changes requested):

clients.asciidoc
amanusk: Good to know, I tried to see if there are any code conventions regarding this but couldn't find any. 
A question, I am trying to reference a paragraph in a different chapter (clients.ascii), this does not seem to work well with regular references. 
Meaning, if I reference something by using `<<refrence>>`, where `[[reference]]` is not in the same file, the reference does not work. 

The workaround I found is to use `<<clients.asciidoc#reference, ParagraphName>>` . Is there a better way to do this, or should I only reference the chapter?

fjrojasgarcia:
Thanks for your contribution @amanusk. Good job!

fjrojasgarcia (approved):

PR #410, opened 2018-04-07
slight reorganization and small changes to clients

aantonop:


clients.asciidoc
```diff
@@ -438,7 +445,21 @@ Cipher Browser:: a full-featured Ethereum-enabled mobile DApp browser and wallet
...
-Some of the more popular ones include MyEtherWallet/MyCrypto and MetaMask.
+Some of the more popular ones include MetaMask, Jaxx and MyEtherWallet/MyCrypto
+
+==== MetaMask
+
+MetaMask was introduced in <<intro>>, and is a versatile browser-based wallet, RPC client, and basic contract explorer. It is available on Chrome, Firefox, Opera and Brave Browser. Find MetaMask at:
+
+https://metamask.io
+
+At first glance, MetaMask is a browser-based wallet. But, unlike other browser wallets, MetaMask injects a web3 instance into the browser, acting as an RPC client that connects to a variety of Ethereum blockchains (eg. mainnet, Ropsten testnet, Kovan testnet, local RPC node, etc.). The ability to inject a web3 instance and act as a gateway to external RPC services, makes MetaMask a very powerful tool for developers and users alike. It can be combined, for example, with MyEtherWallet or MyCrypto, acting as an web3 provider and RPC gateway for those tools.
```
fjrojasgarcia, line 456: For the moment I think we can be satisfied with this approach of the important role played by Metamask, although I also think that sooner rather than later we will have to revisit our definitions and explanations as Metamask evolves as well as our own knowledge of this important tool.

PR #415, opened 2018-04-07
testnets: Added detail, re-organized flow

aantonop:
This one is in progress. Please review but expect more pushes to the branch!


ethereum-testnets.asciidoc
```diff
@@ -1,34 +1,49 @@
...
 
 === Using Testnets
 
-You can either connect to publicly available test networks or spawn a development network of your own. First, let's use a public testnet for easier setup. To use a public testnet requires some testnet ethers and a connection to that network. For testnet ethers, faucets are used. To connect to testnets, we will discuss how to do so in various popular clients.
+You can either connect to publicly available test networks or spawn a private test network of your own. First, let's use a public testnet for easier setup. To use a public testnet requires some testnet ether and a connection to that network. For testnet ether, "faucets" are used, which distribute test ether slowly, "dripping" out small amount to anyone who asks. To connect to a testnet, you need an Ethereum client, either a full client such as geth, or a lightweight client such as MetaMask.
```
rphmeier: calling metamask a "lightweight client" further adds to the confusion people have about what an Ethereum light client is, what the security model is, and how it differs from a full node. For correctness, we would want to call it a thin client, but that's not a very approachable term for a newcomer.

ethereum-testnets.asciidoc
```diff
@@ -1,34 +1,49 @@
...
-https://infura.io/docs/gettingStarted/chooseaNetwork
+When MetaMask connects to a test network, it uses the Infura service provider for the JSON-RPC interface. Infura was born with the goal of delivering stable and reliable RPC access to the internal projects within ConsenSys. In addition to a JSON-RPC API, Infura also offers a REST (Representational State Transfer) API, IPFS (Interplanetary File System, ie. decentralized storage) API, and a Websockets (ie. streaming) API.
+
+Infura offers gateway APIs to the Ethereum main net, Ropsten, Kovan, Rinkeby, and INFURAnet (a custom testnet for Infura).
```
rphmeier: a note about the security implications of a trusted RPC endpoint (on the mainnet especially) would be good; lots of new users wouldn't be aware how much this increases the trust requirement.

ethereum-testnets.asciidoc
fjrojasgarcia: I agree with @rphmeier. For me it is unnatural to call lightweight client to MetaMask, although honestly, I still do not know the proper term. I think we need to think a bit about this point.

In their website they say, "MetaMask is a bridge that allows you to visit the distributed web of tomorrow in your browser today. It allows you to run Ethereum dApps right in your browser without running a full Ethereum node".

AFAIK there is no SPV built into web3.js, and the provider is the absolute source of truth.

People tend to call "lightweight client" to every client that is not a full node. IMHO the concept of "lightweight client" comes from an SPV client that only makes sense in the UTXO model not in the Ethereum account model.
Anyway, surely the broader term "lightweight client" will end up imposing for all the cases.

ethereum-testnets.asciidoc
fjrojasgarcia: :+1:  @rphmeier

ethereum-testnets.asciidoc
```diff
@@ -1,34 +1,49 @@
...
+====
+
 === Connecting to Testnets
 
 ==== Metamask
-Metamask fully supports the Ropsten, Kovan and Rinkeby testnets, but connecting to other testnets and local networks is also possible. In Metamask, simpy clicking the drop down that says 'Main Network' allows for switching networks. MetaMask offers an option to buy ether. If the Ropsten testnet is used, ether can be obtained from the Ropsten Test Faucet Service. You can access this faucet from this page. It requires the Metamask extension to work. https://faucet.metamask.io/
+Metamask fully supports the Ropsten, Kovan and Rinkeby testnets, but connecting to other testnets and local networks is also possible. In Metamask, simpy clicking the drop down that says 'Main Network' allows you to switch networks. MetaMask also offers an option to "buy" test ether,  . If the Ropsten testnet is used, ether can be obtained from the Ropsten Test Faucet Service. You can access this faucet from this page. It requires the Metamask extension to work. https://faucet.metamask.io/
```
fjrojasgarcia: Here **"ether,  . If the Ropsten"** has a comma and a dot, Maybe only the dot?

sgvictorino:
Using PRs for project management is fun! I made some review comments.

ethereum-testnets.asciidoc
rphmeier: The closest analogue to an SPV client in Ethereum is an actual light client, which verifies the header chain and fetches state data (pinned to the merkle root) only as necessary to serve user requests. This is terminology that has been used since well before the launch of the Ethereum mainnet, and it's very different from a "thin" client which is just a shell connected to a trusted remote node. 

The term "thin client" has some historical merit (~20 years or so) in the gamedev industry, where it refers to a multiplayer game client which simply forwards input to a server and displays what it is sent.

PR #429, opened 2018-04-08
Transaction page change adding nonce info about double spends

chanan-sack:


aantonop:
I think you've identified the correct place for the nonce clarification regarding double-spend. However, the grammar and sentence structure is not ideal. I will re-write it a bit.

thoragio:
I've drafted a clarification that you might want to use as a starting point:

> The main purpose of the nonce is to sequence transactions, which prevents double spending. A nonce is a counter whose value is incremented after each transaction, just like a numbered ledger in a bank account. Each value of the nonce can only be used once and its associated debit/credit can only be applied in sequence, therefore maintaining the history of transactions for the originating address.

PR #431, opened 2018-04-08
Definitions for digital signature, faucet, lightweight client, and testnet

ivanmolto:


glossary.asciidoc
```diff
@@ -107,6 +110,9 @@ EVM Assembly Language::
 
 Fallback function::
     It's like a fishing net to catch all the ether that is sent to a contract.
+    
+Faucet::
+  A website that dispenses rewards in the form of free test ether for developers who want to do test on testnets.
```
fjrojasgarcia, line 115: Here we could add ...
"Faucets were used initially as a way to promote and publicize the cryptocurrency of its own network little by little. The main idea was to give away small amounts of cryptocurrency for free to spread the word about the new electronic currency to millions of users".

The first bitcoin faucet was called The Bitcoin Faucet and was developed by Gavin Andresen in 2010.

fjrojasgarcia (approved):

PR #436, opened 2018-04-09
added description of EVM

chanan-sack:
When learning about the EVM the doc jumps directly into comparison without briefing the reader about the subject, so i felt a need to add a basic description

### DIFF
--- a/evm.asciidoc
+++ b/evm.asciidoc
@@ -3,6 +3,7 @@
 
 [[evm_description]]
 === What is it?
+The part of the protocol that actually handles internal state and computation is referred to as the Ethereum Virtual Machine (EVM). From a practical standpoint, the EVM can be thought of as a large decentralized computer containing millions of objects.
 
 [[evm_comparison]]
 ==== Compare to


evm.asciidoc
```diff
@@ -3,6 +3,7 @@
 
 [[evm_description]]
 === What is it?
+The part of the protocol that actually handles internal state and computation is referred to as the Ethereum Virtual Machine (EVM). From a practical standpoint, the EVM can be thought of as a large decentralized computer containing millions of objects.
```
fjrojasgarcia, line 6: I agree that it is better for the reader to start with an introduction than with a comparison. I had that same feeling the first time I read this chapter.

I would even elaborate this introduction.

I would leave the PR open by encouraging you to continue along this line and as long as Andreas is in agreement with it.

fjrojasgarcia (approved):

evm.asciidoc
aantonop: Agreed

chanan-sack: will work on extending the opening paragraphs, thank you both

Issue #446, opened 2018-04-09
Easy editing work: Heading style normalization and standardization 


aantonop:
We need to normalize heading styles across the book:

Please do one pull request PER asciidoc file, to avoid large merges. Adjust heading style in each section as follows:

* Top level heading is level 2 (level 1 is the book title itself). Each asciidoc file should start with a "==" heading.
* All lower case, except for first letter, proper nouns and acronyms. "What is this thing?", "What is the Ethereum sprocket?" "Who created the Ethereum Name Service (ENS)"
* Acronyms are spelled out, capitalized, with the acronym in parentheses. 
* No period at the end. Question mark if it is a question (generally avoid question headings, unless really appropriate)
* Should include a unique anchor (see #279), all lower case, underscore separated. 
* Headings should be followed by a blank line.
* Heading should be followed by a paragraph of text, not a lower-level heading without any text. If you find one like this, add a TODO comment (line of 4 slashes "////", line with "TODO: add paragraph", line of 4 slashes)

fjrojasgarcia:
Should we consider them [contribute's ](https://github.com/ethereumbook/ethereumbook/blob/develop/CONTRIBUTE.md) notes?

NukeManDan:
At this point, think this is closed @aantonop ? close #446 ?

PR #451, opened 2018-04-10
dapps.asciidoc heading style normalization and standardization

nuke-web3:
hope this is what you were looking for in #446 for the dapps.asciidoc file.

Let me know what else I can do for this guy - otherwise I will tackle another chapter in a similar fashion.

### DIFF
--- a/dapps.asciidoc
+++ b/dapps.asciidoc
@@ -1,10 +1,16 @@
+[[decentralized_applications_chap]]
 == Decentralized Applications (DApps)
 
-== What is a DApp?
+////
+TODO: add paragraph
+////
 
-DApp stands for Decentralized Application. Unlike a traditional Application, a DApp does not belong only to one single provider or server, but the whole stack would be deployed and operated in a distributed fashion on a peer-to-peer network.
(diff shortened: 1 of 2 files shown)


nuke-web3:
Also:
multiple level two (==) headings in this doc - should we consider shifting all to higher level beyond title?

abhiShandy:
Only the chapter title should have 2nd level heading. Sections and sub-sections will have 3rd and 4th level headings respectively.

aantonop:
Yes, only the chapter/section should be level 2, everything else should be level 3 and below. 

Also, once you have spelled out an acronym in one heading I think we can keep it as an acronym in subsequent headings (eg. DApp from the second heading forwards). 

Nice work!

nuke-web3:
Thanks @aantonop ! I am about to correct the levels and acronyms - might be a good idea to update your issue to reflect that for others contributing

Elenore23:
> hope this is what you were looking for in #446 for the dapps.asciidoc file.
> 
> Let me know what else I can do for this guy - otherwise I will tackle another chapter in a similar fashion.

PR #471, opened 2018-04-11
contribution notes proposals

fjrojasgarcia:
Contribution notes proposals taken from #446 and others.


nuke-web3:
might consider adding a bit more based on #451 :
--only the chapter/section should be level 2, everything else should be level 3 and below.
-- once you have spelled out an acronym in one heading I think we can keep it as an acronym in subsequent headings

PR #485, opened 2018-04-12
first initial sync, Andreas fav lot dots ;) and gateways typo

manilabay:


clients.asciidoc
```diff
@@ -113,10 +113,10 @@ Syncing the Ethereum blockchain is very input-output (I/O) intensive. It is best
...
-* CPU with 2+ cores
-* Solid State Drive (SSD) with at least 50GB free space
-* 4GB RAM minimum, 8GB+ if you have an HDD and not SSD
-* 8 MBit/sec download Internet service
+* CPU with 2+ cores.
+* Solid State Drive (SSD) with at least 50GB free space.
+* 4GB RAM minimum, 8GB+ if you have an HDD and not SSD.
+* 8 MBit/sec download Internet service.
 
 These are the minimum requirements to sync a full (but pruned) copy of an Ethereum-based blockchain.
```
SylTi, line 121: We must be very carreful here, a fastsync node does not validate the state after every blocks like a bitcoin pruned node would do, in only start something like 1000 blocks before the current state.
Also I'm not sure about the requirement: last week I tryed to sync geth in fast mode, I couldn't even catch up using my dev laptop (i7, 16gb ram) because the i/o are just too much for a hdd
Also I couldn't completly sync and I'm already over 64gb

manilabay: just adding dots in this commit
please my suggestion would be to open an issue if you found something to review

manilabay (commented):
just adding dots in this commit
I suggested open an issue if somebody found something to review

SylTi:
You're write. Opening an issue

clients.asciidoc
aantonop: @SylTi Please add an issue about this. I agree, we need to clarify and address it.

PR #509, opened 2018-04-18
Added oracle example, section on cryptlets

keleffew:
Added oracle example, section on cryptlets, updated numbers for footnote references

### DIFF
--- a/oracles.asciidoc
+++ b/oracles.asciidoc
@@ -69,7 +69,7 @@ Oraclize provides a service that allows decentralized applications to request th
 
 The concept of a 'cryptlet' as a standard for verifiable oracle truths has been formalized as part of Microsoft's wider ESC Framework [7].  Cryptlets execute within an encrypted capsule that abstracts away the infrastructure, such as I/O, and has the CryptoDelegate attached so incoming and outgoing messages are signed, validated, and proven automatically.  Cryptlets support distributed transactions so that contract logic can take on complex multi-step, multi-blockchain and external system transactions in an ACID manner.  This allows developers to create portable, isolated, and private resolutions of the truth for use in smart-contracts.  Cryptlets follow the format below:
 
----
+----
 public class SampleContractCryptlet : Cryptlet 
   { 
         public SampleContractCryptlet(Guid id, Guid bindingId, string name, string address, IContainerServices hostContainer, bool contract)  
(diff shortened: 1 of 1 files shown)


oracles.asciidoc
```diff
@@ -66,7 +67,9 @@ So far, we have only discussed oracles in the context of requesting and deliveri
...
+The concept of a 'cryptlet' as a standard for verifiable oracle truths has been formalized as part of Microsoft's wider ESC Framework [7].  Cryptlets execute within an encrypted capsule that abstracts away the infrastructure, such as I/O, and has the CryptoDelegate attached so incoming and outgoing messages are signed, validated, and proven automatically.  Cryptlets support distributed transactions so that contract logic can take on complex multi-step, multi-blockchain and external system transactions in an ACID manner.  This allows developers to create portable, isolated, and private resolutions of the truth for use in smart-contracts.
+
```
timnugent: It's not clear to me how this fits into the "Computation oracles" section - perhaps this could be expanded upon? "Oracle truths" sounds like data, rather than computation. Might this fit better in the "Data Authentication" section? 

Also:
- What is a CryptoDelegate?
- What is an encrypted capsule? Is this a hardware enclave?
- Has the ACID acronym been introduced yet?
- How do cryptlets differ from the TownCrier approach?

oracles.asciidoc
```diff
@@ -67,7 +67,18 @@ So far, we have only discussed oracles in the context of requesting and deliveri
...
+---
+public class SampleContractCryptlet : Cryptlet 
+  { 
+        public SampleContractCryptlet(Guid id, Guid bindingId, string name, string address, IContainerServices hostContainer, bool contract)  
+            : base(id, bindingId, name, address, hostContainer, contract) 
+        { 
+            MessageApi = 
+                new CryptletMessageApi(GetType().FullName, new SampleContractConstructor()) 
+--- 
+
```
timnugent: I assume this is C# code? How does a smart contract interact with it? I think it might be more useful to show the (e.g. Solidity) interface rather than this. It also seems a little out of place; perhaps both this and the interface could be moved to the "Oracle client interfaces in Solidity" section.

oracles.asciidoc
```diff
@@ -11,6 +11,7 @@ However, many blockchain applications require access to extrinsic information. T
 * Capital markets data: pricing baskets of tokenized assets/securities
 * Benchmark reference data: incorporating interest rates into smart financial derivatives
 * Static/pseudo-static data: security identifiers, country codes, currency codes
+* Time and Interval data: Event triggers grounded in precise SI time measurement
```
timnugent, line 14: Interval --> interval
Event --> event

Perhaps introduce the SI acronym if not already done

PR #521, opened 2018-04-23
Consensus chapter changes

chronaeon:
Added descriptions for _metrics_ (measuring points) around which PoS and PoW chains operate. Metrics can be either risk-based, or hash-bashed, (alternatively work-based.) The former metric relies on network nodes properly identifying and creating risk, while the latter metric relies on network nodes properly identifying and validating a certain scheme of hashes derived from public data. In both cases, the _risk_ or the _hash_ is the particular, identifiable, measuring factor (namely, the metric) by which **consensus** may be achieved on each particular chain.

### DIFF
--- a/consensus.asciidoc
+++ b/consensus.asciidoc
@@ -2,28 +2,48 @@
 
 == Consensus
 
-When it comes to decentralized record keeping, which is how blockchain technology is often described, it becomes infeasible to rely on trust alone to ensure that factual information is added to the record books. This problem arises in decentralized networks as a result of the lack of a central entity that would typically decide what should and should not be considered as fact. It's difficult to avoid this issue when working with blockchain technology because it is this lack of a central entity that is one of the main contributors to blockchain's popularity. In order to circumvent this issue a network must have a system in place by which a consensus can be reached by the nodes in the network.
-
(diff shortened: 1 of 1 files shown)


fjrojasgarcia:
Awesome contribution @chronaeon, Thanks a lot!

consensus.asciidoc
```diff
@@ -2,9 +2,48 @@
...
+
+=== Hash-Based Metrics
+
+Commonly known as Proof-of-Work (PoW) metrics, these metrics establish consensus because protocols that use them set computers to work on finding difficult numbers. For the sake of illustration, consider supercomputers whose only job in life is to search integer-space for prime numbers. Now consider a whole network of average computers. These computers, when put together, might be said to have the combined power of a supercomputer. The only job for this network of computers is similarly to search possible numbers for another kind of number called a SHA-256 hash. These numbers have unique properties, just like prime numbers, that make them easily identifiable when discovered.  
+
+When SHA-256 hashes are computed accurately they serve as an attestation that a certain amount of computing power has been used to find the number. The most recent prime number to be found is asciimath:[2^(77,232,917) − 1]. It was found by a computer just like SHA-256 hashes are found by computers. SHA-256 hashes are easier to find than new prime numbers, however, and this is where hash-based metrics derive their power.
```
fjrojasgarcia, line 16: Are we sure that asciimath is running fine here?

aantonop:
OReilly renders with latexmath extensions (which don't work on Github). I
don't know about asciimath.

PR #525, opened 2018-04-24
clients: fix typo

amanusk:
Small typo fix: whicu -> which


fjrojasgarcia:
;-) Good catch @amanusk!

PR #526, opened 2018-04-24
Adding class inheritance

tpmccallum:
Writing about inheriting pre-written code.
Whilst this is the premise of OOP it also can be dangerous for immutable systems.
Also cover how auditing obscure code and jumping between files is difficult for code auditors and developers alike.


vyper.asciidoc
```diff
@@ -78,7 +78,7 @@ modifier onlyBy(address _account)
...
 
-Class inheritance:: TODO explain class inheritance and then paraphrase the following ... "requires people to jump between multiple files to understand what a program is doing, and understand the rules of precedence in case of conflicts (which class's function X is the one that's actually used?). Hence, it makes code too complicated to understand."
+Class inheritance:: Inheritance allows programmers to harness pre-written code by acquiring pre-existing functionality, properties and behaviors from existing software libraries. Inheritance is powerful and promotes the reuse of code. However, it also creates an issue for those who write and audit code. Inheritance forces programmers and code auditors, who seek to understand precisely what a program is doing, to jump between multiple files while storing complex information in their memory. It can be argued that, given the immutability of blockchains, leveraging powerful, obscure and sometimes unintelligible software libraries can wreak havoc in smart contract implementation. For example a programmer inadvertently triggering pre-written code in an incorrectly coded smart contract.
```
fjrojasgarcia: This point belongs to the principles on which Vyper is based, which on the other hand, **currently could be subject of  debate** in the Ethereum community.

Therefore, IMHO, I am not sure that the book is the most appropriate place to **emphasize** it. I would only mention it **without emphasizing it too much**.

Anyway, let's wait @aantonop express his opinion and decide.

tpmccallum:
Thanks @fjrojasgarcia 
I have reworded to be more specific (related to inheritance) and also provided a link to Solidity documentation which supports the view that multiple inheritance can be problematic.

fjrojasgarcia:
A good attenuation of the emphasis. Thank you very much for listening to my opinion @tpmccallum.
I think that now it has a tone of impartiality that IMHO was necessary.

Issue #313, opened 2018-03-21
Sending value in a contract creation


nicksavers:
It's possible to send a value in contract creation ('_[contract registration](https://github.com/ethereumbook/ethereumbook/blame/324b691e80449f64f63a4a37c3145cf6daafa3f5/transactions.asciidoc#L314
)_'). That value would then be immediately added to the balance of the newly created contract. 

AFAIK, that is disabled by default in Solidity, so in order to receive ether, the fallback function must be marked payable. And make sure you have a way to retrieve the ether, otherwise (with the consensus rules as they are) it will just stay in there forever. http://solidity.readthedocs.io/en/v0.4.21/contracts.html?highlight=creation#fallback-function

See also page 9, paragraph 7 item (78) in the Yellowpaper https://ethereum.github.io/yellowpaper/paper.pdf

![New account with value](https://i.imgur.com/Jz1DDqd.gif)

fjrojasgarcia:
Pointing to your Gitter note --> you can pre-fund an account .... and only retrieve it after contract creation. Nobody would know that there is no private key to the address, just a transaction to create a contract to send the ether value somewhere.

fjrojasgarcia:
More of your awesome clarifications from Gitter --> 
... «Since a contract address is derived from the sender address and nonce, you can know the contract address ahead of time and send ether to it. Then only when you actually send the contract creation transaction, it can be spent. Not before.
To the outside world, until then it would just look like a normal EOA.»

fjrojasgarcia:
The address for an Ethereum contract is deterministically computed from the address of its creator (sender) and how many transactions the creator has sent (nonce). The sender and nonce are RLP encoded and then hashed with Keccak-256.

https://ethereum.stackexchange.com/questions/760/how-is-the-address-of-an-ethereum-contract-computed

starmanzz9:
```
contract DAPP {
    
    function DAPP() payable {
        //accept Ether during deployment
        //..
    }
    
    event Donation(address _addr, uint _val);
    function () public payable { //fallback
        // accept Ether after deployment
        assert(msg.value > 0);
        emit Donation(msg.sender, msg.value);
    } 
}
```
[...]

PR #530, opened 2018-04-25
Add example of publishing a contact from web3

amanusk:
-  Replace `TODO` with an example publishing a contract from web3 interface 

- Add screenshots from etherscan displaying the published contract

- Add contributor


transactions.asciidoc
```diff
@@ -360,6 +356,74 @@ While the zero address is only intended for contract registration, it sometimes
 Any ether sent to the contract registration address +0x0+ or the designated burn address +0x0...dEaD+ above will become unspendable and lost forever.
 ====
 
+A contract registration transaction should contain no ether value, only a data payload that contains the compiled bytecode of the contract. The only effect of this transaction is to register the contract.
```
fjrojasgarcia, line 359: Perhaps for this we need to revisit "Sending value in a contract creation" #313 wisely opened by @nicksavers 
I mean, something like **"should not, but could"**.

fjrojasgarcia (changes requested):
For me a great job, but please revisit "Sending value in a contract creation" #313 and think about my suggestion.

transactions.asciidoc
amanusk: I have not written this line, so I did not want to change it. I have only moved its order. After the original `TODO` there was a paragraph about a specific address for burning ether. I have moved it closer to the explanation about address `0x0` since I thought it will be strange to  return to it after a long example. 

Regardless, I was not able to create the `Faucet.sol` contract with a transaction where `value != 0`

fjrojasgarcia: I see, I got it, sorry and Thanks @amanusk

PR #531, opened 2018-04-26
Updating dapp.tools sections and creating MakerDAO section

apmilen:


dapps.asciidoc
```diff
@@ -164,16 +164,17 @@ Github link; https://github.com/embark-framework/embark
...
-* Package management
-* Source code building
-* Unit testing
-* Simple contract deployments
+* Easily use any version of the C++ Solidity compiler
+* Run unit tests and interactively debug contracts in a native EVM execution environment
+* Create persistent testnets using the Go Ethereum client
+* Easily deploy your dapp to any EVM blockchain
+
+It was created in the spirit of the Unix design philosphy, which means it's a good citizen of the command-line and can be easily composed with other tools. To get started, visit https://dapp.tools/dapp
```
fjrojasgarcia: Did you mean **"philosophy"** rather than **"philosphy"**? Or could it be valid?

dapps.asciidoc
```diff
@@ -164,16 +164,17 @@ Github link; https://github.com/embark-framework/embark
 
 Website link; https://github.com/embark-framework/embark
 
-[[dapp_develotment_tool_sec]
-==== DApp (development tool)
-DApp is a simple command line tool for smart contract development. It supports these common usecases:
+[[dapp_develotment_tool_sec]]
+==== Dapp (development tool)
```
fjrojasgarcia: I thought that there was an explicit section for **dapp as a tool** in the [**dev-tools.asciidoc**](https://github.com/ethereumbook/ethereumbook/blob/develop/dev-tools.asciidoc) chapter, and that this chapter was for Dapp as an App not a tool.
It is also true that it appeared here before this PR. Perhaps @aantonop could give us some light on it.

dapps.asciidoc
```diff
@@ -164,16 +164,17 @@ Github link; https://github.com/embark-framework/embark
 
 Website link; https://github.com/embark-framework/embark
 
-[[dapp_develotment_tool_sec]
-==== DApp (development tool)
-DApp is a simple command line tool for smart contract development. It supports these common usecases:
+[[dapp_develotment_tool_sec]]
```
fjrojasgarcia: I do not know if there is a criterion admitted in the book for the denomination of a **Decentralized App**, I can not say whether we should use **DApp** or **Dapp**.

fjrojasgarcia: Here is a typo **"develotment"** -> **"development"**.

fjrojasgarcia (changes requested):

dapps.asciidoc
apmilen: I agree that its unclear where the dev tools should go. I was just following the existing layout and figured it would get properly placed during the editing phase.

apmilen:
@fjrojasgarcia are there any other changes you think I should make for this PR

PR #533, opened 2018-04-26
comments what is ethereum chapter

phiuofficial:
Some suggestions on how to make this intro chapter more complete, biggest of which is the explanation of Ether (now only gas is covered).

I used the 4/ to insert the comments, I hope that this is the format you would expect, if not please advise.

As a side comment for the rest of the book, I read few chapters and I had the feeling that sometimes the authors seem to be shy talking about bitcoin directly (instead referring to hypothetical blockchains). It feels like someone else wrote this, not Andreas.


phiuofficial:
please ignore the consensus file, it’s a duplicate I didn’t manage to get rid of. Thanks for your understanding

fjrojasgarcia:
To allow **automatic merge** processes, it is preferable that _PRs_ be published with the **changes that are explicitly proposed**.

IMHO, manually ignoring changes to a PR is outside the philosophy of **git** as an advanced code management tool.

My recommendation would be to **create a work branch based on a synchronized version of the branch "develop"** and move to it the explicit changes that are proposed in this PR. Then cancel this PR and publish the new one from the new created work branch.

You could follow the comments from @iinstances in **Gitter** _(abr. 26 09:44)_.

A little effort today could be a skill tomorrow.

:smiley:

PR #534, opened 2018-04-26
Included section on estimating gas.

naveensrinivasan:


gas.asciidoc
```diff
@@ -67,11 +67,26 @@ In which case:
...
+Source: https://ethereum.stackexchange.com/a/515/19763 
+License: CC0 
+Added by: @naveensrinivasan 
 ////
 
+Estimating gas works by pretending the transaction was actually being included in the blockchain, and then returning the exact gas amount that would have been charged if that pretend operation was real. In other words, it uses the exact same procedure a miner would use to calculate the actual fee but never mined into the blockchain.
+
+Note that the estimate may be significantly more than the amount of gas actually used by the transaction, for a variety of reasons including EVM mechanics and node performance.
+
+``` javascript
```
fjrojasgarcia, line 82: Only to point out to [gas estimation for a contract method with Contract abstraction](https://github.com/ethereumbook/ethereumbook/commit/65dcb31a5aa93ff3a408c29bb23718cb746adabb#diff-55565c7afc8cd081294729f4c9e4fc97);

```
var contract = web3.eth.contract(abi).at(address);
var gasEstimate = contract.myAweSomeMethod.estimateGas(arg1, arg2, {from: account});

```

naveensrinivasan: @fjrojasgarcia I don't understand your comment. Are you recommending that I change the code to point to the other reference?

fjrojasgarcia: Yes, I'd suggest that it be offered to the reader to jump to another part of the book where the estimation of gas is made with an alternative method based on the abstraction of contract, but I think it could be for another iteration.

PR #536, opened 2018-04-27
dev-tools: emerald and sputnikvm

splix:


fjrojasgarcia:
Super interesting contribution @splix, Thanks a lot!

PR #543, opened 2018-04-29
Tokens: spelling corrections, grammar condsiderations.

dcheli:
spelling corrections, and suggestion to add Utility to list in "How Tokens are used" because utility tokens are discussed later in the chapter but not specifically named in the list.


fjrojasgarcia:
Hi @dcheli and thanks for the catchs!

PR #553, opened 2018-05-01
1st section of consensus chapter

Orlandao:
I have expanded out this introductory paragraph to better contextualize the challenge of consensus within decentralized organizations. This is my first commit so please let me know if this kind of editing is useful.  :) Orlando


aantonop:
Nice work, well written. Thanks!

PR #566, opened 2018-05-06
Remove duplicated tip in ethereum-testnets

lautarodragan:


fjrojasgarcia:
Good catch and Thanks for your first contributiion @lautarodragan!

PR #571, opened 2018-05-11
Error in oraclize snippet

Fakje:
The oraclize query callback function should only be callable by oraclize. As-is in the code snippet, the function is callable by anyone but oraclize.


Fakje:
Also, should not assert be replaced by require ?

timnugent:
Thanks - will merge this for now. re: assert/require, Thomas Bertani (Oraclize) is going to replace this with a better example anyway.

PR #597, opened 2018-05-28
fix a typo patter -> pattern

cockroachzl:


fjrojasgarcia:
Good catch @cockroachzl!
Thanks for your contribution.

PR #598, opened 2018-05-28
fix typo

h0ngcha0:
Fix type:
infromation -> information


fjrojasgarcia:
Good catch @liuhongchao!
Thanks for your contribution.

PR #603, opened 2018-05-28
adding missing semicolons in the code snippet

cockroachzl:
to correct the syntax of event declaration


fjrojasgarcia:
Good catch @cockroachzl!
Thanks for your contribution.

PR #607, opened 2018-05-29
Capitalize Bitcoin where appropriate in wallets.asciidoc

corysolovewicz:
https://bitcoin.org/en/vocabulary#bitcoin

> Bitcoin - with capitalization, is used when describing the concept of Bitcoin, or the entire network itself. e.g. "I was learning about the Bitcoin protocol today."
> 
> bitcoin - without capitalization, is used to describe bitcoins as a unit of account. e.g. "I sent ten bitcoins today."; it is also often abbreviated BTC or XBT.


fjrojasgarcia:
Totally agree @corysolovewicz.
Thanks for your contribution!

PR #614, opened 2018-05-31
Removed duplicate POW subheader

masterlook:
Removed a duplicate POW subheader


fjrojasgarcia:
Thanks for your contribution @masterlook.

PR #617, opened 2018-05-31
Updated consensus for consistency

masterlook:
Updated all entries of the following acronyms for consistency:
POW -> PoW
POS -> PoS
POA -> PoA
DPOS -> DPoS


fjrojasgarcia:
Absolutely agree @masterlook

PR #618, opened 2018-05-31
fix URLs to use https

gene-smith:


corysolovewicz:
Nice catch!

gene-smith:
changing http to https

PR #619, opened 2018-05-31
Consensus: Filled in section on why GPU

masterlook:
Added some content for the section discussing why use GPUs for mining.

### DIFF
--- a/consensus.asciidoc
+++ b/consensus.asciidoc
@@ -113,7 +113,13 @@ The *PoW algorithm* involves: +
 
 ==== PoW Function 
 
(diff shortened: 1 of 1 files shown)


fjrojasgarcia:
**Just as a reminder**;

**The current status of the book is CONTENT FREEZE** - TECH REVIEW. Only errors, fixes and typos will be accepted.

Any new content submissions will be delayed for the second edition, which goes into production in late 2019 or early 2020.

masterlook:
ok thanks for the reminder.

PR #626, opened 2018-06-02
Changed event to events

kennethhoytwoodruff:


fjrojasgarcia:
Good catch @kennethhoytwoodruff!

PR #631, opened 2018-06-05
Typo in smart-contracts.asciidoc

tomas-aftalion:
Not an expert but seems like a typo: SELFDESTRUCT for SEFDESTRUCT.


fjrojasgarcia:
Yes it's a typo. good catch @tomasaftalion!
Thanks for your contribution.

PR #634, opened 2018-06-07
reference code/truffle/FaucetEvents

tomas-aftalion:
There is no code/Solidity/FaucetEvents


fjrojasgarcia:
Good catch @tomasaftalion!

PR #638, opened 2018-06-10
Fixed typos in clients.asciidoc

HusamElbashir:
Parity was written from the "ground up" in Rust -> Parity was written "from the ground up" in Rust
chainging -> changing
RCP -> RPC


fjrojasgarcia:
Good catchs @HusamIbrahim!

PR #658, opened 2018-07-11
Fixed many typos (like it's -> its), removed UTF-8 apostrophes and other fixes.

Jo-Con-El:
I know that the guidelines specify 1 PR per file, but I'm not sure if this diff qualifies as it's fixing many typos across the board.  If you still prefer that, I can split the changes in several diffs.


aantonop:
Please rebase on develop, too many conflicts from tech-review

PR #721, opened 2018-07-27
Added compiler specification in Faucet.sol

mazewoods:


aantonop:
This version of Faucet.sol is the example code right *before* the concept of pragma directives is introduced. We add the version pragma in Faucet1.sol. It is puposefuly missing from Faucet.sol

mazewoods:
Clear, thank you for clarifying that up.

PR #726, opened 2018-07-29
Fix maths display for markdown

WizardOfAus:
Fixed: 

[latexmath]
++++
\begin{equation}
{y^2 = (x^3 + 7)}~\text{over}~(\mathbb{F}_p)
\end{equation}
++++

To display properly with y^2^ = x^3^ + 7


aantonop:
While latexmath is not supported by the Github rendering engine, it is supported by the O'Reilly Atlas rendering engine. These equations look correct in the printed book and ebook.

PR #728, opened 2018-07-30
Added pros and cons for PoW and PoA mechanisms (plus minor grammar fixes)

auino:


aantonop:
What you term "disadvantages" are often seen as advantages by PoW proponents. I'd rather keep things more neutral, by reviewing differences without labeling them as advantages or disadvantages.

PR #773, opened 2018-08-23
Work on DApps chapter

rrthomas:


rrthomas:
Not sure exactly what happened here: @tpmccallum's recent commit went straight to the `tech-review2` branch, and when I simply rebased on `origin/develop` git complained that I was behind `origin/tech-review2`, and then when I rebased on `origin/tech-review2` I got all the

PR #787, opened 2018-09-04
Grammar and tool pass on chapter 4

morenoh149:
* fix grammar in 2 sentences
* update output of `helpeth` to reflect the output from the latest version of the tool


morenoh149:
btw for `As we will see in more detail below, an Ethereum address for an EOA is generated from the public key portion of a PC key pair.` should `PC key pair` be `EC key pair`?

PR #794, opened 2018-09-17
OpenZeppelin GitHub URL returning a 404

tpmccallum:
Updating the URL from 
https://github.com/OpenZeppelin/zeppelin-solidity/blob/master/contracts/token/ERC20/StandardToken.sol
which appears to have moved to
https://github.com/OpenZeppelin/openzeppelin-solidity/blob/master/contracts/token/ERC20/ERC20.sol


fjrojasgarcia:
As always, thanks a lot for the catch @tpmccallum!

come-maiz:
it's a bad idea to link to master on any repo. You should use the latest released tag, which should always remain valid:
https://github.com/ethereumbook/ethereumbook/pull/808

PR #795, opened 2018-09-17
fix grammar and punctuation in preface

dkarvounis:
"communicate" should be "communicates".
"relay on" should be "relay".
Remove extra comma after "thus".


fjrojasgarcia:
Thanks for your first contributiion @dkarvounis!

PR #796, opened 2018-09-17
remove redunant text

dkarvounis:
Remove the third sentence of the MEW section, which is redundant with the first.


fjrojasgarcia:
Good catch again @dkarvounis!

PR #802, opened 2018-09-20
Updating URLs

tpmccallum:
Getting this message when following the Ian Coleman BIP39 link
"This page has moved.
You will be redirected in 5s to https://iancoleman.io/bip39/"
I have updated the source asciidoc accordingly

### DIFF
--- a/wallets.asciidoc
+++ b/wallets.asciidoc
@@ -297,7 +297,7 @@ There is also a BIP-39 generator implemented in a standalone webpage, which is e
 .A BIP-39 generator as a standalone web page
 image::images/bip39_web.png["BIP-39 generator web-page"]
 
-The page (https://iancoleman.github.io/bip39/) can be used offline in a browser, or accessed online.
+The page (https://iancoleman.io/bip39/) can be used offline in a browser, or accessed online.
 
 [[create_hd_wallet]]
 ==== Creating an HD Wallet from the Seed


fjrojasgarcia:
Given the **current status** of the project, I have doubts about whether or not it is time to proceed with **merges**, even though these seem urgent.

Although **your PRs are all an important contribution**, I would prefer the opinion of @aantonop before continuing to perform merges.

**Let's try not to lose coordination with him**, since it is a critical point in the good achievement of the project.

**Serve this comment, for the rest of current and next PRs**, until knowing the opinion of @aantonop .

Thanks a lot @tpmccallum!

aantonop:
All these changes will have to be applied manually (by me) after copy edit
and before final proof, if the copy editor doesn't catch the same errors.

Please submit PRs for errors (like this link) that a non-domain-expert copy
editor might miss. Do not submit PRs for typos, grammar and syntax, as
those are part of the copy editors job

Thanks

fjrojasgarcia:
In the sake of clarification I have added the section **Current Project STATUS** to **CONTRIBUTING.md** and also the points to keep in mind, in case that the PROJECT STATUS is CONTENT FREEZE, in the section **Contributing with a Pull Request** of the same file.

I hope you are in agreement with this change and that in turn it helps to improve the coordination among contributors.

@aantonop , thanks and apologies if we have not been totally alienated with your criteria with the last merges.

tpmccallum:
Thanks @fjrojasgarcia 
Sounds great @aantonop 
The "Files changed" tab in #801 and #802 should be all you need to make the manual URL changes (if required at all; as you say copyediting may resolve these).

Also, just to let you know I have now written a script which follows the public links in the asciidoc files. Aside from #801 and #802 there is only one other link which needs fixing. I have created a PR #803 because I believe, as you suggested, that perhaps this one may throw a non-domain-expert copy editor.

Please let me know if you want the link checking script for future use; it's just too hard to manually do thinks like link checking right? ;-)

fjrojasgarcia:
@tpmccallum, that sounds great to me also.

IMHO, _"./tools"_ seems to be a good place to put these kind of stuff.

Anyway, will be better listen to @aantonop about it.

PR #808, opened 2018-09-29
Update the openzeppelin link to the latest release

come-maiz:
Things in master are always changing a lot. Always use a tagged version. The latest as of today is v1.12.0


aantonop:
Leo, having difficulty with the openzeppelin changes and we're in the last
stage of copy edit. I sent a message on openzeppelin gitter, can you take a
look? The METoken doesn't compile...

come-maiz:
@aantonop yes, we just read your message. I'm searching for other links that point to our master, to tag them. Let us know if there's something else we can do to help.

PR #813, opened 2018-10-01
Complete node command and further instructions about where to get infura api key 

karolinkas:
Add a bit of further instructions about where to get infura api key and complete node command

- if you are a newcomer to node it might not be totally obvious how to run these commands
- explain briefly how the infura api key is being used


karolinkas:
Alt Text Looking for more!

PR #814, opened 2018-10-01
Small grammar and typo fixes

Jawknee-88:


Jawknee-88:
Ah, only just noticed @aantonop mentioned these sorts of changes will be handled by the publisher. Feel free to reject.

PR #826, opened 2018-10-23
fix: Change ETC wording to reflect it being original version

YazzyYaz:
The wording about ETC in the `contrib` directory were inaccurate. ETC historically is the original chain, not a forked chain.


phyro (commented):
Another small thing that confused me: `Project known as` = `smart contract known as` is perhaps more informative/accurate

PR #833, opened 2018-11-02
Adding compiler version for Remix IDE

yerinle:
- Tried running the Faucet.sol contract on Remix IDE, got a compiler error message that was resolved by adding this line


aantonop:
Current project status is CONTENT FREEZE - IN PRODUCTION.

The Faucet.sol example is purposely missing pragma, as that command is introduced with Faucet1.sol

Issue #834, opened 2018-11-05
Typo in Chapter 3


francoabaroa:
First off, thanks so much for this book! It's extremely helpful.

Second, I want to confirm this typo in the following section:

https://github.com/ethereumbook/ethereumbook/blob/develop/03clients.asciidoc#the-json-rpc-interface

The text I am referring to is the following: 

_The response, 0x430e23400, tells us that the current gas price is 1.8 gwei (gigawei or billion wei). If, like us, you don’t think in hexadecimal, you can convert it to decimal on the command line with a little bash-fu:_

```
$ echo $((0x430e23400))

18000000000
```

Should it be 18 gwei instead of 1.8 gwei in the above text?

I will open up a PR

PR #835, opened 2018-11-05
[nit] Fix typo for gwei representation

francoabaroa:
Issue here: https://github.com/ethereumbook/ethereumbook/issues/834

18000000000 wei should be 18 gwei instead of 1.8 gwei


wbnns:
@francoabaroa 

Thank you for the PR. It has already been resolved on `develop`.

PR #838, opened 2018-11-11
Added missing links to other chapters and some external documentation

ghost:


0xmountaintop (changes requested):
I suggest using  relative paths instead of absolute paths, so that we can be independent on github.

wbnns:
Let's leave this as-is for now, so as to avoid issues with the print version.

PR #844, opened 2018-11-21
Update README.md

johnalanwoods:
Added clear instructions with regard to how to read the book, similar to the bitcoinbook.


README.md
```diff
@@ -13,6 +13,10 @@ This repository contains the draft of the book, currently in progress, scheduled
 
 There is no PDF version available. While it is possible to produce one, and you may do so for personal use, please **DO NOT** publish a PDF. The publishers and the authors provide a very generous open license that allows anyone to read this book for free on GitHub. It is not convenient to read that way, but it is free. Convenience, in the form of a nicely packaged PDF, ebook or printed book, is what the publishers charge money for and what pays for the publishers and authors work. Just like buying bottled water instead of drinking from the tap, buying a printed copy, PDF or ebook pays for the packaging. If you publish a PDF copy, you are violating the license, as a PDF is a derivative form and the -ND part of the license means "no derivatives". Doing so makes publishers less likely to publish with open licenses.
 
+## Reading this book
+
+To read this book, see [book.asciidoc](https://github.com/ethereumbook/ethereumbook/blob/develop/book.asciidoc). Click on each of the chapters to read in your browser. Other parties may choose to release PDFs of the book online.
```
rattle99, line 18:
```suggestion
To read this book, see [book.asciidoc](https://github.com/ethereumbook/ethereumbook/blob/develop/book.asciidoc). Click on each of the chapters to read in your browser.
```

johnalanwoods: I think releasing a PDF is against the license.

rattle99: > I think releasing a PDF is against the license.

Yes, you are right. That's why I thought maybe removing that line would be better since it kind of encourages releasing PDFs.

wbnns: Let's leave it as-is, to match, for now.

PR #848, opened 2018-11-27
Simple modification

hyunblnjeong:
separate -> separated :)


wbnns:
@gusqls1603 

Thank you. Let's leave as-is. In this case, `separate` is used as an adjective (i.e. distinct, unique) like in the second sentence of the previous paragraph.

PR #857, opened 2018-12-15
Update 09smart-contracts-security.asciidoc to use msg.value not msg.sender

edposnak:
The calculation here should use msg.value not msg.sender


wbnns:
@edposnak Thank you.

PR #860, opened 2018-12-29
Develop

anuphunt:


wbnns:
@anuphunt Thank you, however, let's leave this as-is for now so as to avoid issues with the print version.

PR #861, opened 2019-01-02
Small bug - found selfdestrunct (with an n)

tpmccallum:
Changing selfdestrunct(__recipient_address__) to selfdestruct(__recipient_address__) 
Removing the letter n from the selfdestruct function call on line 341


bgaughran:
Well spotted

PR #866, opened 2019-01-10
Chapter 9: Fix Solidity indentation and spacing

RichHorrocks:
Includes the following changes:
 - Set Solidity indentation to 4 spaces in example code only. This doesn't include code from contracts that are being referenced from elsewhere (e.g. the Parity Multisig contract, which is in a different style to the book's own code.)

 - Changed some spacing above/below comments and around the first lines of contracts/functions to make more consistent.


aantonop:
Spacing changes can mess up the layout in print, where margins are very strict. Please separate the spacing changes from the rest of these changes. I can't accept spacing changes unless I check with the publisher and copy-editors.

PR #869, opened 2019-01-16
Chapter 6: Fixes missing closing tag for u1 var

dbachko:
Before:
screen shot 2019-01-16 at 12 39 00 am

After:
Screen Shot 2020-01-18 at 8 14 36 PM

### DIFF
--- a/06transactions.asciidoc
+++ b/06transactions.asciidoc
@@ -602,7 +602,7 @@ Verification is the inverse of the signature generation function, using the _r_ 
 2. Calculate _w_ = __s__^-1^ _mod p_
 3. Calculate _u~1~_ = __Keccak256__(__m__) * _w_ _mod p_
 4. Calculate _u~2~_ = _r_ * _w_ _mod p_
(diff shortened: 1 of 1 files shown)


HAOYUatHZ:
I suggest that we should keep the symbol *, to be consistent with other parts in the book.

But this PR makes a great point: there is an unexpected underscore in the formula.

0xmountaintop (changes requested):
I suggest that we should keep the symbol *, to be consistent with other parts in the book. But this PR makes a great point: there is an unexpected underscore in the formula.

dbachko:
@HAOYUatHZ fixed

ChrisLinn:
> @HAOYUatHZ fixed

@dbachko would be better if also updating the pic in the PR description

0xmountaintop:
> @HAOYUatHZ fixed

@dbachko would be better if also updating the pic in the PR description

0xmountaintop (approved):

0xmountaintop:
LGTM

wbnns:
@dbachko @HAOYUatHZ Thank you.

PR #870, opened 2019-01-16
Chapter 5: updates asciidoc escaping to allow showing actual character instead of the encoded version

cleathers:
Character encodings were showing in the part of the chapter about HD wallet paths.

This update makes it so when you view the asciidoc on github, you see the wallet path with `'` characters instead of `&#x27;`


wbnns:
@cleathers Thank you. Let's leave this as-is for now, so as to avoid issues in the print version.

PR #871, opened 2019-01-19
Remove faulty &thinsp; from ethereum foundation address

lukasz-golebiewski:


wbnns:
@lukasz-golebiewski Thank you.

PR #875, opened 2019-01-30
Update 09smart-contracts-security.asciidoc

jbazzel:
The stated goal of the "Owner Operations" section is to show how, if the owner with elevated security loses their key, the entire token becomes inoperable. The token IS inoperable but not for the intended reason. The double equals on line 1979 means that isFinalized can never be set to true. The token is inoperable whether the owner has their key or not.


wbnns (approved):
@jbazzel Thank you.

PR #879, opened 2019-02-11
[type] gasPrice * gas fee/gas limit

nakajo2011:
It maybe miss type.


wbnns:
@nakajo2011 Thank you.

PR #885, opened 2019-03-02
Online Vyper compilers

tpmccallum:
There has been some discussion in the community about finding consistent compiler versions online. I learned that there are now actually 3 different online compilers (with different features) available.
I felt that this was important to put in ASAP because having to install the Vyper compiler from source is evidently hindering new adoption of Vyper.


wbnns:
@tpmccallum Thank you.

PR #891, opened 2019-04-03
Chapter 5: Fix prime denotation

robinpan1:
replace &#x27; with '


wbnns:
@robinpan1 Thank you. Let's leave the encoding as-is; this convention is used throughout the book.

PR #894, opened 2019-04-13
Fix the small issue of explaiing IBAN-compatible Direct encoded ICAP address 

milochen0418:
Because 2^156 > 36^30 > 2^155, so we can use 155 bits of Ethereum address when we use Direct ICAP encoding method that is valid in IBAN format. 
 
So the first byte cannot only be a zero byte. The first byte is look like 00000xxxx. 
The first byte just need to smaller than 8.

### DIFF
--- a/04keys-addresses.asciidoc
+++ b/04keys-addresses.asciidoc
@@ -522,7 +522,7 @@ If our address did not start with a zero, it would be encoded with the Basic enc
 
 [TIP]
 ====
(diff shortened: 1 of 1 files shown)


wbnns:
@milochen0418 

Thank you. Let's leave the current explanation as-is. The suggested revision seems a bit more difficult to parse and comprehend. That said, if other contributors disagree we can reconsider.

PR #895, opened 2019-04-13
Chapter 06: Adding missing web3

iphelix:
Missing web3 object when calling getTransactionReceipt function.


wbnns:
@iphelix Thank you.

PR #908, opened 2019-06-09
grammatical error --  09smart-contracts-security.asciidoc

Nirma:
I thought this word ordering was a bit odd after reading this section. 
How about flipping the position of "to"?


wbnns:
@Nirma Thank you.

PR #914, opened 2019-09-12
Homogenise leading whitespace in Solidity examples

Code0x58:
Switch from mixed tabs+spaces in the examples to just spaces


wbnns (approved):
@Code0x58 

Thank you.

PR #941, opened 2020-01-28
ch4 - Added instruction to install helpeth

safetyth1rd:
Added command to install helpeth, as well as advice about npm.

Thanks @wbnns for editing advice and code snippet.


wbnns:
@safetyth1rd Thanks for your help.

PR #945, opened 2020-02-08
fix "Principles of Consensus"

0xmountaintop:
I am not sure what it means by "change the future" in the original text.

From my perspective, future is not yet coming and remains unknow. How can someone manae to change it?

And moreover, "finality" is about whether a history block can be revoked. It looks more like "change the past" to me.


wbnns:
Changing the future is referring to how and what _future_ actions an actor would need to take to in order to revert or modify what previously occurred in the chain. Hope that helps clarify.

Let's leave this as-is for now, but we can reopen if others also feel further elaboration here is needed.

PR #947, opened 2020-02-09
fix step count in "Vickrey auctions"

0xmountaintop:


wbnns (approved):
@HAOYUatHZ 

Thank you.

PR #931, opened 2020-01-25
style: lint contract codes in "code/Solidity"

0xmountaintop:
1. function order should be: constructor -> external -> public -> internal -> private
2. Contracts and libraries should be named using the CapWords style. 
3. 'function () ' -> 'receive()' in solidity 0.6
4. constructor visibility should be explicit
5. The `address` type was split into `address` and `address payable`
6.  It is now necessary to explicitly convert values of contract type to addresses before using an address member. Example: if c is a contract, change c.transfer(...) to address(c).transfer(...), and c.balance to address(c).balance.

PR #949, opened 2020-02-14
Update Faucet.sol to solidity v0.6

safetyth1rd:
-pragma ^0.6 for solc forward compatibility
-fallback function should now be receive() 
-function order should be: constructor -> external -> public -> internal -> private

Changes based on @HAOYUatHZ Pull request #931

### DIFF
--- a/code/Solidity/Faucet.sol
+++ b/code/Solidity/Faucet.sol
@@ -1,9 +1,12 @@
 // Version of Solidity compiler this program was written for
-pragma solidity 0.5.12;
+pragma solidity ^0.6;
 
 // Our first contract is a faucet!
 contract Faucet {
 
+    // Accept any incoming amount
+    receive () external payable {}
+    
     // Give out ether to anyone who asks
     function withdraw(uint withdraw_amount) public {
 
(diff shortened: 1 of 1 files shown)


code/Solidity/Faucet.sol
```diff
@@ -1,9 +1,12 @@
 // Version of Solidity compiler this program was written for
-pragma solidity 0.5.12;
+pragma solidity ^0.6;
 
 // Our first contract is a faucet!
 contract Faucet {
```
0xmountaintop, line 6: I think, can remove the blank line(https://github.com/ethereumbook/ethereumbook/pull/949/files#diff-710fa94076c56efb45ef7449f5d627faL6).
Because there is no such an extra line break in https://solidity.readthedocs.io/en/v0.6.2/solidity-by-example.html#voting
and https://solidity.readthedocs.io/en/v0.6.2/style-guide.html

code/Solidity/Faucet.sol
```diff
@@ -14,7 +17,4 @@ contract Faucet {
         msg.sender.transfer(withdraw_amount);
     }
```
0xmountaintop, line 19: ditto

code/Solidity/Faucet.sol
safetyth1rd: Hey chris, that sounds great. go ahead and make those changes

PR #954, opened 2020-03-11
Typo in Block Header box

0xalizk:
missing "h" in Parent Hash inside Block Header box top-right


wbnns (approved):
@aliatiia 

Nice catch, thanks for fixing.

PR #956, opened 2020-03-21
Correcting information about global state changes

Luker501:
You were incorrectly stating that if a transaction is successful (processed without errors) that all the global state changes it performed are accepted. This is not true when a transaction is sent from an EOA to a contract A, and that contract A calls another contract B using a Solidity method such as .call(...), which will return false on error. And this error could occur after some variables of B were changed. See https://github.com/Luker501/SmartContractInteractions for more information.


lukerQuant:
Still no reviewer assigned to this? It would seem to be a fairly critical piece of information on smart contracts that is currently left out of the textbook

wbnns (approved):
@Luker501 

Thank you.

PR #962, opened 2020-04-06
[chap07] update solc & code to solidity 0.6

0xmountaintop:


adrianmcli:
Just FYI, 0.6 is a massive change and many of the tooling has not caught up to it yet. In fact, many in the community believe it should be called 1.0 instead.

I do think we eventually need to update the code but I am hesitant to endorse the next edition (which is due to come out very soon) should adopt 0.6. However, I am happy to have my mind changed.

HAOYUatHZ:
@adrianmcli I agree with you.  there are a lot breaking changes in 0.6.

I update this chapter to 0.6  because @wbnns suggested me upgrade the whole book to 0.6  ( in Feb).   I am not sure whether you two had later discussions on the upgrading,  but I am happy to keep contributing,  no matter   keeping it as 0.4,  or waiting for 1.0, or so forth.  :D

0xmountaintop:
> Just FYI, 0.6 is a massive change and many of the tooling has not caught up to it yet. In fact, many in the community believe it should be called 1.0 instead.
> 
> I do think we eventually need to update the code but I am hesitant to endorse the next edition (which is due to come out very soon) should adopt 0.6. However, I am happy to have my mind changed.

@adrianmcli I agree with you. there are a lot of breaking changes in 0.6.

I update this chapter to 0.6 because @wbnns suggested me upgrade the whole book to 0.6 ( in Feb). I am not sure whether you two had later discussions on the upgrading, but I am happy to keep contributing, no matter keeping it as 0.4, or waiting for 1.0, or so forth. :D

adrianmcli:
@HAOYUatHZ thanks for the reply. In that case, perhaps I'll defer to @wbnns as well.

I just wanted to voice my opinion that a lot of us working professionally developing dapps or devtools have held off moving to 0.6 because of its breaking changes. I'm not saying that we won't get there, I'm just saying we're not there yet. But perhaps it's fine and having 0.6 in this book will move things along faster in the industry as well. Of course, you guys are the official editors of the book and I will defer to your decision :)

wbnns:
@adrianmcli @HAOYUatHZ 

Thanks for the feedback! Yes, it's a non-trivial amount of work to update and test all of the code. The issue from my perspective is that since this is a new edition, a reader may expect the content to be updated / in-line with the versioning at the time of print/release.

I could imagine there might potentially be some disappointment if one were to get a newly released edition of the book and see a significant amount of code that's out of date. Hence, the recommendation to update it.

@aantonop What are your thoughts?

adrianmcli:
Thinking about this a bit more, I think I'd be in favour of moving to 0.6, provided that it wouldn't take too long to sufficiently run and test all the code to make sure it works. That's something we at Truffle definitely want to help with (we don't want people to have a bad dev experience while reading the book). @HAOYUatHZ how much time do you think it would take to do that?

0xmountaintop:
> Thinking about this a bit more, I think I'd be in favour of moving to 0.6, provided that it wouldn't take too long to sufficiently run and test all the code to make sure it works. That's something we at Truffle definitely want to help with (we don't want people to have a bad dev experience while reading the book). @HAOYUatHZ how much time do you think it would take to do that?

updating the code will be easy, I already have experience in term of that, see https://github.com/ethereumbook/ethereumbook/pull/931 and PR 929~938 (I temporarily close them because I want then queue after 931 and don't really want the PR page looks so messy.)

I have a minor concern that, some of the text may also need to be updated.
For example, this PR removes "constant" function type, we may need to explain it's obsolete in 0.6 otherwise people may get confused if comparing to 0.4.

I can imagine we need to talk about the breaking changes 0.5 and 0.6 introduce, see:

+ https://solidity.readthedocs.io/en/v0.6.0/050-breaking-changes.html
+ https://solidity.readthedocs.io/en/v0.6.0/060-breaking-changes.html

This kind of problem may exist in every possible corner so some chapters need to be checked carefully.

wbnns:
@HAOYUatHZ @adrianmcli 

I connected w/ Andreas. Just to confirm, we should update and test all of the code, and plan to double-check all of the corresponding/supplemental text.

Thank you both for the help!

adrianmcli:
Happy to help.

0xmountaintop:
> @HAOYUatHZ @adrianmcli
> 
> I connected w/ Andreas. Just to confirm, we should update and test all of the code, and plan to double-check all of the corresponding/supplemental text.
> 
> Thank you both for the help!

No problem!

0xmountaintop:
How is everything scheduled? what is the plan?

wbnns:
@adrianmcli @HAOYUatHZ 

Thanks! Regarding timeline, not sure just yet.

Let's see what we can accomplish over the next few weeks.

chanhosuh:
Solidity is now on 0.8, so I hope moving to 0.6 isn't as big a deal.  ;)

How's work going with double-checking all the changes and the text?  I'm helping run a meetup and I thought it might be good for the people learning Solidity to go through the recent version of the text.  Perhaps we could lend a hand if needed?

wbnns:
@chanhosuh

Hi there! All of the code is still in the process of / needs to be updated. That would be great if you all would like to assist with double-checking and updating the samples and other excerpts.

wbnns (approved):

onemliii (commented):
Önemlim

PR #970, opened 2020-05-23
Fix typo from ERC59 to EIP-59

linki:
The text references EIP-59 but then calls it "ERC59" in the link's title. I think it should be "EIP-59" instead.


wbnns (approved):
@linki 

Thank you.

PR #975, opened 2020-07-11
Chapter 7: Fix reference to Event name

henriquegaia:
The correct event name, as seen in https://github.com/ethereumbook/ethereumbook/blob/develop/code/truffle/CallExamples/contracts/CallExamples.sol, is 'callEvent', not 'calledEvent'.


wbnns (approved):
@henriquegaia 

Thank you.

PR #976, opened 2020-07-15
change account[n] to accounts[n] chapter10

jjmr007:
web3 uses `accounts[n]` (instead of `account[n]`), to select an element of the `accounts` array. Little mistakes in paragraphs 461, 472 & 496 (as far as I find them).


wbnns (approved):
@jjmr007 

Thank you.

PR #977, opened 2020-07-16
Update 02intro.asciidoc

MentatX:
Small fix for chapters links.


wbnns (commented):
@MentatX

Hello, thanks for the suggestion, however, I think this would contradict the convention that's followed throughout the book for cross-linking. If you're having a problem with this link in the version that you're reading, please let us know.

PR #978, opened 2020-07-20
Update 02intro.asciidoc: Conclusions links

MentatX:
Fix `Conclusions` chapters links (with  the convention that's followed the book for cross-linking).


wbnns:
@MentatX 

Thanks for the suggestion, but let's leave as-is. The current syntax should work fine and is also used (e.g. `git grep 'smart_contracts_chapter'`).

Hope that helps clarify.

PR #985, opened 2020-08-05
Update Faucet8.sol

carlinhocr:
Modify constructor and this object to support current default truffle configuration (compiler 0.5)


wbnns (approved):
@carlinhocr 

Thank you.

PR #988, opened 2020-08-16
Chapter 4: update big number comparison

wtong98:
The comparison of 10^77 possible keys to 10^80 atoms in the universe implies that there is 1 key per 10^80 / 10^77 = 1000 atoms, which is not quite "almost enough private keys to give every atom" an account. Instead, I've narrowed the scope to just Earth (and sand), which may be more numerically correct.


aantonop:
The number of atoms in the observable universe (Eddington Number N_edd_) is
estimated between 10^77 and 10^80. Therefore, at the lower range is exactly
1 key per atom. Perhaps instead of changing the comparison, we clarify the
range of Nedd

wbnns:
Pushed 201e814 to clarify range.

PR #998, opened 2020-10-22
Added references to multiple node providers to the web3js Appendix tutorial.

deric-alchemy:
Rewrote some of the `web3js` demo code to enable multiple node provider options.


deric-alchemy:
@aantonop Please take a look here!

PR #1009, opened 2021-01-26
Fix undeclared identifiers

harrigan:
The `balance` and `transferValue` variables in the `TimeLock` and `EtherGame` contracts are undeclared. Also, I renamed the `balance` variables to `transferValue` in the `TimeLock` contract to avoid confusion with the `this.balance` variable and for consistency.


wbnns (approved):
@harrigan 

Thank you.

PR #1011, opened 2021-02-06
fix the "TimeLock.sol" sample contract

FeiyangTan:
Missing variable declaration, added missing variable "balance" in the sample contract "TimeLock.sol"


wbnns (approved):
@FeiyangTan 

Thank you.

PR #1026, opened 2021-04-25
Fix constant width example styling in preface

nickcruz:
Seems like constant width needed to use back ticks to show up well in preview mode.


wbnns:
@nickcruz 

Heya! Thanks for the suggestion. It seems this would make the Constant width become indistinguishable from Constant width bold, which could be problematic:

image

PR #1031, opened 2021-06-01
remove minor asciidoctor warning

lvnilesh:
- expecting 1 received 7
 - expecting 2 received 8
 - expecting 3 received 9


wbnns:
@lvnilesh 

Hello, thanks for your PR! The previous numbering is correct, however, because it is a continuation of the previous ordered list:

image

Hope that helps clarify! 👍

PR #1035, opened 2021-07-13
Function modifiers separated by spaces

delaaxe:
Unless I misunderstood something, function modifiers are separated by spaces, and not commas?


wbnns (approved):
@axeldelamarre 

Nice catch, thanks for fixing this!

PR #1036, opened 2021-07-13
Correct use of "function" when it should be "variable"

jdmuriel:


wbnns (approved):
@jdmuriel 

Thanks for fixing! 👍

jonesk7734 (approved):
😀

PR #1037, opened 2021-07-29
Update raw_tx_demo.js for new lib versions

sipox11:
As of 30/07/2021, copy pasting this code directly didn't work for me. I was getting this exception: 

.../node_modules/ethereumjs-util/dist/bytes.js:80
```throw new Error("Cannot convert string to buffer. toBuffer only supports 0x-prefixed hex strings and this string was given: " + v);```

This commit barely sets the data field to 0x0 in order for the script to work. 

Test OS: macOS
Node version: v12.18.4
Ethereumjs-tx version: 2.1.2


amadisme:
I have the same problem.Did you fix it?

pangsirirat422 (approved):

PR #1094, opened 2021-11-01
Updated 02intro.asciidoc with an alternative faucet

sudotman:
MetaMask test faucet tends to glitch out a lot of times and not send any ether. An alternative faucet would be helpful.


sudotman:
has been merged into the main branch through their automated bot

PR #1109, opened 2022-01-28
Grammar fix in 07smart-contracts-solidity

liamjsc:
Fixing subject-verb agreement in point (4) "...in a manner that does not propagate..."


dav3yblaz3 (approved):
this should work

Issue #1116, opened 2022-02-19
broken link


php4fan:
At https://github.com/ethereumbook/ethereumbook/blob/develop/07smart-contracts-solidity.asciidoc#writing-a-simple-solidity-program

> In [[intro_chapter]](https://github.com/ethereumbook/ethereumbook/blob/develop/07smart-contracts-solidity.asciidoc#intro_chapter), we wrote our first Solidity program

That [intro_chapter] links to https://github.com/ethereumbook/ethereumbook/blob/develop/07smart-contracts-solidity.asciidoc#intro_chapter which does not exist.

hassan-stash:
@php4fan Hi, I've raised this PR for the issue #1116  Appreciate your review. Thanks

aantonop:
Links refer to anchors that may be in other chapters and are correctly
referenced when the book is compiled into PDF or other formats. They are
not "valid" in Github viewer only.

adityakode:
@php4fan  I think the issue is solved . Ps check once. If you feel the issue is solved, you may close the issue

PR #1122, opened 2022-03-13
fix for broken link #1116

hassan-stash:
ref the reported issue #1116 
linked reference to the first solidity program (Faucet contract example) in the intro chapter .


aantonop:
Links in asciidoc can refer to other chapters because they are meant to be compiled into a book, not viewed only as separate files. This "fix" would break the book compiler.

PR #1126, opened 2022-04-05
Gas Consideration for Testnet

ololade97:
A testnet now deducts a transaction fee, that's gas. A free ether will not be sent to another testnet address without gas.


dav3yblaz3 (approved):
see if this improves things for dev front end

dav3yblaz3 (approved):

PR #1135, opened 2022-05-30
Marking the sender address as 'payable'

kirby88:
This is not compiling with latest version of solidity https://github.com/ethereumbook/ethereumbook/blob/develop/02intro.asciidoc#a-simple-contract-a-test-ether-faucet


kirby88:
This was not compiling https://github.com/ethereumbook/ethereumbook/blob/develop/02intro.asciidoc#a-simple-contract-a-test-ether-faucet

dav3yblaz3 (approved):

PR #1142, opened 2022-07-27
Update 01what-is.asciidoc => fix broken links on Further Reading

foufrix:
fix broken links :
- Merkle Patricia trees
- Ethash PoW algorithm

Note that  Ethereum Virtual Machine list of resources: https://eth.wiki/en/concepts/evm/ethereum-virtual-machine-(evm)-awesome-list is also broken but not sure what it was referring to?
Maybe this is an equivalent: https://ethereum.org/en/developers/docs/programming-languages/golang/#other-aggregated-lists

### DIFF
--- a/01what-is.asciidoc
+++ b/01what-is.asciidoc
@@ -152,7 +152,7 @@ https://github.com/chronaeon/beigepaper
 https://github.com/ethereum/devp2p/blob/master/rlpx.md
 
 * Ethereum Virtual Machine list of resources:
-https://eth.wiki/en/concepts/evm/ethereum-virtual-machine-(evm)-awesome-list
+https://ethereum.org/en/developers/docs/evm/
 
 * LevelDB database (used most often to store the local copy of the blockchain):
 https://github.com/google/leveldb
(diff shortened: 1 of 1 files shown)


haurog:
I just fired up the waybackmachine to look at the broken link. This is a snapshot from October 2020:
https://web.archive.org/web/20201027031126/https://eth.wiki/en/concepts/evm/ethereum-virtual-machine-(evm)-awesome-list
This website seems to be a copy of: https://github.com/pirapira/awesome-ethereum-virtual-machine
So you could link to the github page, but it was last updated 4 years ago, so is probably a bit dated.

foufrix:
@haurog good idea to use a web archive. From what you said, it looks like what I send in my message seems more up-to-date though

haurog:
@foufrix thanks for fixing the links. The links in the pull request are perfect replacements for the broken ones. 

I had a look at the old EVM content and your suggestions. Yes, your suggestion is definitely more up to date. Unfortunately, what you would link to is rather for solidity developers coming from the golang (GO) programming language. So it links to resources how to handle smart contracts and presents various projects using the go programming language in the Ethereum space.

The original (now broken) link however is focussed on the EVM. This more or less the virtual machine were the smart contracts are running on. 

So, your suggested link is newer, but it does not really fit the need to get more information about the EVM. Luckily the ethereum.org page actually has a chapter on the EVM: 

https://ethereum.org/en/developers/docs/evm/

This is a description of the EVM which is actually better than the now broken link. So I would replace the broken link with the one above.

foufrix:
Updated with what you advised :)

Issue #1138, opened 2022-07-05
How to compile Faucet.sol in remix?


wudapeng268:
I compile Faucet.sol in remix, but got following error. It seems version error. Does anyone know how to fix it?

`
ParserError: Source file requires different compiler version (current compiler is 0.8.7+commit.e28d00a7.Emscripten.clang) - note that nightly builds are considered to be strictly less than the released version
--> xx.sol:4:1:
|
4 | pragma solidity ^0.6.4;
| ^^^^^^^^^^^^^^^^^^^^^^^
`

dicethedev:
I think you should change your solidity version to the updated one. [ pragma solidity ^0.8.7; ] and check whether is going to fix the error.

abdurahman-harouat:
use compiler 0.6.0 and

```
pragma solidity 0.6.0;
```

PR #1145, opened 2022-09-04
How to compile Faucet.sol in remix? #1138

AnandK-2024:
### merge this updated file #1138 
# change solidity version and update function of incoming account


AnandK-2024:
update faucet.sol

Adamsbite:
Guess this is aamazig

PR #1146, opened 2022-09-14
Develop

lilsunny243:


SerafDosSantos:
Why not take fr_CA?

aantonop:
We can add any languages we want. I'll add fr-CA

PR #1154, opened 2022-10-18
added "that offers a complete guide"

BirdboyBolu:
I think this improvement helps users see the total value of what they will get from the book.


BirdboyBolu:
made a few changes to the sentences.

PR #1165, opened 2023-02-20
Correct link to EVM chapter

luloxi:


aantonop:
The links are made to work for the print edition (where all chapters are
joined together), not for GitHub. This breaks the print edition, sorry.

PR #1196, opened 2023-11-03
Text editing for clarity and updating the consensus algorithm to PoS

mcmoodoo:


mcmoodoo:
Who are the reviewers/verifiers? Would appreciate some ownership here to drive our mutual efforts in making this resource better!

Thanks.

PR #1269, opened 2025-12-01
Inaccurate description of bilinear pairing in KZG

DarkArtistry:
T should be described as the target group of the bilinear pairing, not as an elliptic curve.

In BLS12-381/KZG, G_1 and G_2 are elliptic-curve groups; G_T lives in a finite field extension.


alessandromazza98 (approved):
thanks, you're right🙏

PR #1296, opened 2026-04-01
Fixed Syntax Error in Chapter 6 

groverInnovate:
await should be inside a async function but for fetching nonce it was defined outside async function 

nonce: await provider.getTransactionCount(wallet.address)


alessandromazza98:
there's no need of such modification

PR #1300, opened 2026-04-17
Change balances and withdrawalLimit types to int256.

spyrosblock:
The previous code was not expoitable. If the balance is of type uint256, then when the attacker tries to exploit it, the line `balances[msg.sender] -= _amt;` will cause an underflow exception.

You can look at https://github.com/spyrosblock/reentrancy-demo

### DIFF
--- a/src/chapter_9.md
+++ b/src/chapter_9.md
@@ -50,21 +50,21 @@ Reentrancy can be tricky to grasp without a practical example. Take a look at th
 
 ```solidity
 1 contract EtherStore {
-2      uint256 public withdrawalLimit = 1 ether;
+2      int256 public withdrawalLimit = 1 ether;
 3      mapping(address => uint256) public lastWithdrawTime;
-4    mapping(address => uint256) balances;
+4    mapping(address => int256) balances;
 5
 6    function depositFunds() public payable{
-7      balances[msg.sender] += msg.value;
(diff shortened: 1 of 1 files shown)


alessandromazza98 (changes requested):
this is actually true only if you compile this contract with Solidity 0.8+. Previous solidity versions just wraps around, they don't automatically check underflow and revert.

I'd leave it as it is because this example is just to show how reentrancy works, we don't really care about all implementation details.

Thank you nevertheless for your effort 🙏 Eventually in the future open an issue before the PR so that we can give you a feedback before you put work into it.

spyrosblock:
Thank you for your response. The reason I didn't create an issue was that this example confused me and I couldn't understand if the attack would work or not. So I just tried it and thought that it was wrong. So I just created the pull request with the code that is vulnerable. I think it might be better to have a comment or something so that people don't get confused.

alessandromazza98:
> Thank you for your response. The reason I didn't create an issue was that this example confused me and I couldn't understand if the attack would work or not. So I just tried it and thought that it was wrong. So I just created the pull request with the code that is vulnerable. I think it might be better to have a comment or something so that people don't get confused.

good point. If you wanna create the PR with the comment I'm down. I can also do it

spyrosblock:
Ok. I will create a PR.

PR #1309, opened 2026-04-21
fix: multiple modifiers are applied as a space-separated list not comma-separated

### LINKED ISSUE #1308: Multiple modifiers are not a comma-separated list
In

https://github.com/ethereumbook/ethereumbook/edit/develop/src/chapter_7.md

it says that:

```
More than one modifier can be applied to a function; they are applied in the sequence they are declared, as a comma-separated list.
```

I think the last part should be "as a space-separated list" or something like that because they are not comma-separated.

Should I create a PR?

spyrosblock:
Closes #1308


alessandromazza98 (approved):
thank you ❤️

PR #1310, opened 2026-04-22
improve wording on BLS aggregation verification in ch4

Oscarzhangtc:
fix incorrect grammar/diction that can confuse the reader


alessandromazza98 (approved):
Makes sense thanks

PR #1316, opened 2026-05-12
Fix Chapter 14: Correct exponent notation

0x76agabond:
This PR fixes incorrectly rendered exponent notation in Chapter 14:
- `2256` -> `2^256`
- `00` -> `0^0`


alessandromazza98 (approved):
thanks

PR #1320, opened 2026-08-20
Update JSON-RPC request description for clarity

Chengxuan:
Thanks for the great book! ❤️ 

A minor wording suggestion. Update the description of the JSON-RPC request object: use the term "member" consistently, as per the JSON-RPC spec and the bullet point sections below.

It caused confusion for me when I was reading the description of `id` below.


alessandromazza98 (approved):
thank you

PR #1322, opened 2026-09-18
Correct the ERC-20 zero-value transfer example

pashap9990:
Chapter 10 incorrectly states that ERC-20 leaves zero-value transfer behavior unspecified.

The ERC-20 specification requires zero-value transfers to be treated as normal transfers and to emit the Transfer event. This applies to both transfer and transferFrom. A transfer may fail for other validation reasons, but zero amount alone is not a valid reason to revert.

Reference: ERC-20 specification (https://eips.ethereum.org/EIPS/eip-20#transfer)


alessandromazza98 (approved):
thanks 🙏

PR #1323, opened 2026-09-18
Correct the gas payment explanation

pashap9990:
Chapter 6 incorrectly says gas is deducted only after transaction execution.

EIP-1559 charges the sender for the maximum gas amount before execution, then refunds the unused portion afterward. The sender must also have enough ether to cover the transaction value.

Gas is paid upfront. The EIP-1559 reference implementation (https://eips.ethereum.org/EIPS/eip-1559#specification) deducts `gas_limit × effective_gas_price` before calling execute_transaction, then refunds unused gas afterward.

This PR updates the paragraph to explain that process clearly and accurately.


alessandromazza98 (approved):
lgtm, thanks

PR #1324, opened 2026-09-19
Correct assert gas behavior in Chapter 7

pashap9990:
> “assert behaves differently from require when it fails: it consumes all remaining gas.”

  This is outdated. Since Solidity 0.8.0, a failed assertion uses REVERT instead of INVALID, preserving unused gas. The book explicitly targets Solidity 0.8.26.


alessandromazza98 (approved):
correct thanks

PR #1325, opened 2026-09-19
Fix account nonce explanation in Chapter 6

pashap9990:
Correct the claim that nonces are calculated from transaction history. Explain that the nonce is stored in account state and increments even when transaction execution reverts. Note that EIP-7702 authorization processing can also increment it.


src/chapter_6.md
```diff
@@ -200,7 +200,7 @@ The *nonce* is one of the most important and least understood components of a tr
 
 > Nonce: A scalar value equal to the number of transactions sent from this address or, in the case of accounts with associated code, the number of contract-creations made by this account.
 
-Strictly speaking, the nonce is an attribute of the originating address—that is, it only has meaning in the context of the sending address. However, the nonce is not stored explicitly as part of an account's state on the blockchain. Instead, it is calculated dynamically by counting the number of confirmed transactions that have originated from an address.
+Strictly speaking, the nonce is an attribute of the originating address—that is, it only has meaning in the context of the sending address. However, The nonce is stored explicitly as part of the account’s state. For an externally owned account, it increments whenever a transaction from that account is included in the blockchain, even if execution reverts. EIP-7702 authorization processing can also increment it.
```
alessandromazza98:
```suggestion
Strictly speaking, the nonce is an attribute of the originating address—that is, it only has meaning in the context of the sending address. However, the nonce is stored explicitly as part of the account’s state. For an externally owned account (EOA), it increments whenever a transaction from that account is included in the blockchain, even if execution reverts. EIP-7702 authorization processing can also increment it.
```

alessandromazza98 (approved):
Makes sense thanks.

PR #1326, opened 2026-09-20
Correct the explanation of ERC-20 transferFrom and permit

pashap9990:
> “Both require the sender to interact directly with the blockchain…”

This is false for transferFrom. The token owner signs an approve transaction, but the spender or authorized contract submits the later transferFrom transaction. The owner does not need to interact directly for that transfer.


alessandromazza98 (approved):
makes sense. I don't think the old sentence is wrong, but this clarifies it

